PR #242, opened 2014-08-18
Avoid trunks too if "avoid_motorway" is selected.

aceman444:
Trunks are already considered as motorways in the "prefer_motorway" option below so make it consistent. This should fix ticket 811, which is now inaccessible to me due to unknown reasons.


aceman444:
OsmAnd now avoids motorroad=yes together with motorways, so that would be a solution for our country. But this is another case where per-country defaults in OsmAnd would be useful. There are already many of them.

PR #280, opened 2014-11-28
Suppressing warning in ttsconfig.p

gorn:
There were several warnings in swi-prolog when reading the file. I have suppressed the warnings which are not oging to be corrected and corrected others. I am not familiar with turbo prolog, so please test it. According to their manual it should work. If this patch get thtough, I can prepare similar patches to other ttsconfig files, so there are not warnings and it is little cleaner.


vshcherb:
Not so important for now

PR #290, opened 2015-01-26
Update routing.xml; Refine penalties

hvdwolf:
Add this moment there are no penalties when going from motorway to motorway_link (and trunk to trunk_link). It sometimes happens that when a motorway_link exiting from a motorway, running parallel to the motorway, and leading back to the motorway is chosen by OsmAnd. This happens when the motorway_link is the inner bend and the motorway is the outer bend, causing the motorway_link to be 1-2 seconds faster (in theory).
The other requirement is that both the motorway and motorway_link don't have a maxspeed or have the same maxspeed.

Additionally I lowered the overall select value as in cities the value of 250 is a bit too much. For longer routes it is less critical.

Solution for penalties contributed by Peter B. in the mailing list.
Lowering overall value based on an issue in the mailing list and by a little trial & error by myself.

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -158,16 +158,16 @@
 			</if>
 			
 			<select value="0" t="highway" v="motorway"/>
-			<select value="0" t="highway" v="motorway_link"/>
-			<select value="0" t="highway" v="trunk"/>
-			<select value="0" t="highway" v="trunk_link"/>
+			<select value="10" t="highway" v="motorway_link"/>
+			<select value="10" t="highway" v="trunk"/>
+			<select value="15" t="highway" v="trunk_link"/>
 			<select value="50" t="highway" v="primary"/>
 			<select value="50" t="highway" v="primary_link"/>
 			<select value="100" t="highway" v="secondary"/>
 			<select value="100" t="highway" v="secondary_link"/>
 			<select value="150" t="highway" v="tertiary"/>
 			<select value="150" t="highway" v="tertiary_link"/>
-			<select value="250"/>
+			<select value="200"/>
 			
 			<!--
 			<select value="" t="highway" v="road"/>


vshcherb:
I found the same issue but in that case I used priority to lower it (I think my change was recent). Could you please check using your test case that you have the issue which is reproducible without this fix. May be it was fixed by already

hvdwolf:
I checked my problem area (the one I could remember) and that one works
fine.

I will ask in the mailing list.

However, My issue that requires the lowering of the overal value of 250 is
still not solved.

Harry

2015-01-26 22:14 GMT+01:00 vshcherb <notifications@github.com>:

hvdwolf:
Peter B tested also.
He used your routing.xml with your modified priorities and that one works OK.
He also tested "my" routing.xml with the heuristic coefficient of 1.35 for car routing. That resulted for him in a few motorway_link deviations. Then he added to my "1.35" routing.xml the speed penalties and then it works OK again.

Both your routing.xml and my  "1.35" routing.xml work fine in the two occasions where I had the issues (I found the second one back). If I add Peters penalties both files still work OK for my situations.


In short: your routing.xml works OK.
My "triple-fast" "1.35" routing.xml needs Peters speed penalties for Peters situations, but not for my situations.

Finally: the overall value of 250 for the speed penalties is still to high for a number of situations in my tests.

vshcherb:
I feel the pull-request is ok. But I will merge it a bit later, so we can still continue discussion, what will be the best way to organize test framework. Any idea? How could we share test success & test failures.

Also I'm curious did you try anything with bicycle routing?

hvdwolf:
I have not looked into the test framework yet. I did see it is possible from OsmandMapCreator but did not spend any time on it yet.
I have a set of "Favorites" for which I know these might cause problems (and also added now the forgotten second set). These favorites are from experience during travelling or "playing from the sofa".

I have been using the bike routing during summer but was not too enthusiastic about it. Brouter does indeed function better but not perfectly either. Of course also the sometimes incomplete tagging in OSM regarding bicycle options criples any bicycle routing.
In the last wek I switched to creating a GPX route in Routeconverter and use that one in OsmAnd.
Currently, in winter, I don't use it at all but I'm eagerly following the "Unpaved roads" mail conversation currently going on (https://groups.google.com/forum/#!topic/osmand/QSCY6OpA-Gg).
I'm planning to spend more time on bicycle routing again when the sun starts to shine again in spring.
At this moment it is not really a priority to me.

PR #291, opened 2015-01-27
Add "abutters" tag to routing section.

aceman444:
Roads with "abutters" set may indicate whether the road is in a city (place) or not. Basing e.g. maxspeed on this tag can be beneficial for some countries.

And the "toll" rule seemed to be duplicate so I remove it.


vshcherb:
I don't think you need it for car. High level means distances > 200km (shorter 50km high level not used at all), but it is not recommended to overload it with lots of data

aceman444:
Sorry, I didn't understand your comment. What did you want to say?

vshcherb:
I don't think you need it for car - adding abbuters base=true
It is used only for high level navigation (base=true attributes). 
 High level means distances > 200km (shorter 50km high level not used at all), but it is not recommended to overload it with lots of data

So base=true is just very few tags which could impact on navigation > 200 km

aceman444:
Ok. It seems the latest maps generated with this new tag are affected by using the abutters tag in the routing.xml file, meaning I get different time estimates when using it.
So all seems to be working, thanks.

PR #294, opened 2015-01-28
Update ispeech_ko.csv

swkim01:


sonora:
I think this is generated automatically from ttsconfig.p

PR #295, opened 2015-02-03
Update routing.xml

Glebzilla:
When you use "Avoid unpaved" at bicycle routing, there is no use of "smoothness" tag. At this time there are 260 000 + using of this tag according to http://taginfo.openstreetmap.org/keys/smoothness

So I added this tag to bicycle routing to \if param="avoid_unpaved"> with values very_bad, horrible, very_horrible, imapassable (according to wiki) and made values same as tracktype (0.4, 0.1, 0.1, 0.1). Also I added "smoothness" to \way attribute="priority">.

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -432,6 +432,10 @@
 				<select value="0.1" t="tracktype" v="grade3"/>
 				<select value="0.1" t="tracktype" v="grade4"/>
 				<select value="0.1" t="tracktype" v="grade5"/>
+				<select value="0.4" t="smoothness" v="very_bad"/>
+				<select value="0.1" t="smoothness" v="horrible"/>
+				<select value="0.1" t="smoothness" v="very_horrible"/>
+				<select value="0.1" t="smoothness" v="impassable"/>
 				<select value="0.1" t="surface" v="unpaved"/>
 				<select value="0.4" t="surface" v="compacted"/>
 				<select value="0.1" t="surface" v="dirt"/>
@@ -450,6 +454,8 @@
 			<select value="0.8" t="surface" v="grass"/>
 			<select value="0.8" t="tracktype" v="grade4"/>
 			<select value="0.8" t="tracktype" v="grade5"/>
+			<select value="0.8" t="smoothness" v="very_horrible"/>
+			<select value="0.8" t="smoothness" v="impassable"/>
 			<select value="0.1" t="access" v="private"/>
 			<select value="0.1" t="barrier" v="debris"/>
 


xmd5a2:
Just to note: according to wiki tracktype is not exactly equals to smoothness. AFAIK tracktype is combination of surface and smoothness.

vshcherb:
The drawback of this approach if 2 tags will be present: trackType and smoothness, the coef could be 0.01 = 0.1 * 0.1 which in my opinion is too low and doesn't reflect reality. Unfortunately I don't see any solution yet in the engine, except we check what combination of tags exist

Glebzilla:
Can we use "if" inside "if"? For example:
<if param="avoid_unpaved">
 	<select value="0.4" t="tracktype" v="grade2"/>
 	<select value="0.4" t="smoothness" v="very_bad"/>
 	 	<if t="trackType" v="grade2">
 		 	<select value="1" t="smoothness" v="very_bad"/>
 	 	</if>
</if>
Or something else.

aceman444:
I would also like to be able to test combinations of tags and boolean operators (tag1 AND tag2, tag1 OR tag2), but it would probably kill the simplicity of the routing file syntax. And the number of combinations that we put into the file would explode.

With the current "first line that matches wins and exits the block" can't you just put the highest penalty to the top? Like:
smoothness=horrible value=0.1
smoothness=very_bad value=0.1
track_type=grade5 value=0.1
track_type=grade4 value=0.1
smoothness=intermediate value=0.4
track_type=grade3 value=0.4
etc...

vshcherb:
If inside if is not supported.
Of course it is not clear even how to define and maintain (may be it will be complex as rendering style).
So it comes to main point, it will be very slow to evaluate for every road segment!

On the other hand there is a solution which is working in progress. Tag transform on server side! So we could combine 2 tags smoothness and track_type into 1, please consult with xmda5 about concrete use cases (we started using it for rendering styles already and it is going to be extended)

xmd5a2:
Tracktype is NOT equals to smoothness. See wiki http://wiki.openstreetmap.org/wiki/Key:tracktype

vshcherb:
The point to do tag transformation at server side, may be introduce tag smoothnes_track_type and do if logic based on it. So instead of if { if else {  } } we can buidl in if (smoothnes_track_type ="bad_track2") else if () elseif ...

aceman444:
Yes, that could also work, but the transformations need to be documented well and accessible. Otherwise people (like me) will put real tags into rounting.xml and wonder why they are not working (because the maps will not contain the real tags but transformed ones, for some cases).
Please send it to the forum once something is done. Not everybody tracks pull requests for each of osmand repositories. Thanks.

PR #297, opened 2015-02-18
Add tourism=apartment to rendering_types.xml

aceman444:
http://wiki.openstreetmap.org/wiki/Tag:tourism%3Dapartment . Is it also necessary to assign an icon to it in some other file?


xmd5a2:
I think new icon is needed.

aceman444:
Could we use one of the existing icons? In which file is this specified?

xmd5a2:
Yes, but new icon is preferable. It is in default.render.xml.

aceman444:
But I can't create an icon. Can you do it?

aceman444:
I've created pll request https://github.com/osmandapp/OsmAnd-resources/pull/298 with a same icon for now.

xmd5a2:
Yes, I'll do it later

PR #299, opened 2015-02-20
Add shop=wine to rendering

aceman444:


xmd5a2:
https://github.com/osmandapp/OsmAnd-resources/blob/master/poi/poi_types.xml and https://github.com/osmandapp/Osmand/blob/master/OsmAnd/res/values/phrases.xml must also be updated. poi_types.xml is our new POI config, phrases.xml is translation config.

aceman444:
Thanks for the hint, those are some new things. I'll add those changes to this pull request.
How do the files relate to specialphrases.txt ? And where are the translations?

xmd5a2:
specialphrases does not used anymore. All translations are done with Weblate. See values-* here https://github.com/osmandapp/Osmand/tree/master/OsmAnd/res

aceman444:
Now that would be fine, finally translators would know which pois are recognized and need translation. I had to guess at that. Hopefully all the existing translations were migrated to the values* file. It would be bad if we had to translate all again. But I do not see the shops' names in https://github.com/osmandapp/Osmand/blob/master/OsmAnd/res/values-sk/strings.xml (e.g. there is no "alcohol"). Do I need to create the phrases.xml file for SK from the english file?

xmd5a2:
No, phrases.xml are automatically updated when translation is done using Weblate. Do not confuse strings.xml and phrases.xml. If there is no slovak phrases.xml that means no translation is done. And yes, translators had to translate again. But I manually converted specialphrases of several common languages to phrases.xml. I can't be done automatically because there was total disorder in internal names and many translations does not worked as intended. https://hosted.weblate.org/projects/osmand/phrases/

aceman444:
OK, I will look at that, thanks. There are some shop values that aren't even defined on the wiki and some english names for some shops are even misleading. Then it is hard to translate them. E.g. "bedding" or "dry goods (sewing)". I'll create a pull request for some.

aceman444:
Will you delete the specialphrases_*.txt files once it is done on weblate? So that users are not mislead. I have done SK version now so speciaphrases_sk.txt could be removed.

xmd5a2:
Yes, we'll delete it after some time

PR #301, opened 2015-03-21
Insert lcn=yes

desperateUser:
localcyclenetwork is very often taggedd with lcn=yes. f.e. http://overpass-turbo.eu/s/8kH.  But it is not able to render it.


xmd5a2:
Merged manually https://github.com/osmandapp/OsmAnd-resources/commit/287cfc5c41659626d0875b26155a507018f741cd

PR #302, opened 2015-03-27
Improve access rules

Zahnstocher:
Proposal for:
https://code.google.com/p/osmand/issues/detail?id=2566

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -92,18 +92,34 @@
 				<select value="-1" t="surface" v="sand"/>
 				<select value="-1" t="surface" v="wood"/>
 			</if>
+
 			<select value="-1" t="motorcar" v="no"/>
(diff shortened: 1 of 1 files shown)


Zahnstocher:
> If you add e.g. `"<select value="1" t="access" v="yes"/>"` there, it will probably allow routing over any type of object even one not on the list of known highways. So e.g. some areas (like park, amenities, buildings), that people marked access=yes. I am not sure we want that.

According to rendering_types.xml parks, amenitys and buildings are not in the routing section (routing_type) of obf file, so they are not routable at all.
Even if access=yes is set on these objects, they are not routable.

vshcherb:
Thanks, that's much better and fix the issue mentionned, but may be not perfect yet )

aceman444:
Ah, half of the logic is in the .obf files again :(

PR #312, opened 2015-05-07
Fix for "Can't find poi typre for poi reference 'town_hall'

mircozorzo:
There was a misspelling town_hall istead of townhall.


mircozorzo:
This change solves also offline map not displayed in nightly build of these days.

PR #313, opened 2015-05-08
Add more weight options in routing.xml

aceman444:
I propose to add more weight options in routing.xml. Until there is a proper support for buses and hgvs in a special routing profile, I think these options need to be in car profile.

I wonder when HGV profile gets supported, it already can be enabled via developer plugin in OsmAnd. But there seem no special settings for it in routing.xml. I'd be happy if OsmAnd got a HGV and BUS profiles. Some of the maxheight, maxweight, maxlength and access options could be moved there so that the car profile does not have too many unnecessary options.


aceman444:
Of course, those HGV and BUS profiles could be made so that they automatically inherit all of what car profile contains, just add some more rules. It would be best if they didn't duplicate all the rules.

vshcherb:
Indeed the car settings should be used, but this is longer goal. So far we are not sure we want to pull it from development settings

aceman444:
Where is the proper place to discuss what is missing to support HGV or BUS routing? I would be interested in that.

Thanks for merging the patch for now.

PR #315, opened 2015-05-10
osmc stripe icons

fred702:


xmd5a2:
It is already done by me but not committed yet.

fred702:
Thanks for your efforts :)

PR #316, opened 2015-05-12
Add support for maxheight in routing.xml

aceman444:
So let's support maxheight in car routing. Do not allow a road that cannot accomodate a car with the user chosen height.
I hope the fractional values will work for numeric parameters. I use type=length as discussed in pull request #1188.


aceman444:
I mean https://github.com/osmandapp/Osmand/pull/1188 .

PR #317, opened 2015-05-13
Fix duplicate attributes in rendering_types.xml

aceman444:
I am not sure what to do with the duplicate to_tag2 at line 1111 (wikipedia at aerodrome). Should it be to_tag3?

### DIFF
--- a/obf_creation/rendering_types.xml
+++ b/obf_creation/rendering_types.xml
@@ -1649,7 +1649,7 @@
 		<type tag="landuse" value="farm" minzoom="6" order="60" />
 		<type tag="landuse" value="farmland" minzoom="6" order="60" />
 		<type tag="landuse" value="farmyard" minzoom="6" poi_with_name="true" poi_category="man_made" />
-		<type tag="landuse" value="forest" minzoom="6" order="58" poi_with_name="true" poi_with_name="true" poi_category="man_made"/>
+		<type tag="landuse" value="forest" minzoom="6" order="58" poi_with_name="true" poi_category="man_made"/>
 		<type tag="landuse" value="garages" minzoom="12" order="71" poi_with_name="true" poi_category="man_made"/>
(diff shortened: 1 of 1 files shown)


obf_creation/rendering_types.xml
```diff
@@ -1649,7 +1649,7 @@
 		<type tag="landuse" value="farm" minzoom="6" order="60" />
 		<type tag="landuse" value="farmland" minzoom="6" order="60" />
 		<type tag="landuse" value="farmyard" minzoom="6" poi_with_name="true" poi_category="man_made" />
-		<type tag="landuse" value="forest" minzoom="6" order="58" poi_with_name="true" poi_with_name="true" poi_category="man_made"/>
+		<type tag="landuse" value="forest" minzoom="6" order="58" poi_with_name="true" poi_category="man_made"/>
```
xmd5a2: I do not agree with this change because almost always forests are junk POIs

obf_creation/rendering_types.xml
```diff
@@ -1747,7 +1747,7 @@
 		<type tag="reef" value="coral" minzoom="11" additional="true" />
 		<type tag="reef" value="rock" minzoom="11" additional="true" />
 		<type tag="reef" value="stone" minzoom="11" additional="true" />
-		<entity_convert pattern="tag_transform" from_tag="natural" from_value="lake" to_tag1="natural" to_value1="water"/>
+		<entity_convert pattern="tag_transform" from_tag="natural" from_value="lake" to_tag1="natural" to_value1="water" to_tag2="water" to_value2="lake"/>
```
xmd5a2: water=lake is not defined as <type> so this change does nothing. But it's ok.

xmd5a2:
Corrected line 1111 by myself

aceman444:
Thanks.

PR #319, opened 2015-05-13
Remove duplicate routing rule in rendering_types.xml

aceman444:
I think the "access" rule seems to be duplicate. The "toll_booth" key rule seems strange as there are no serious usages of such key in OSM data.
I add tourist_bus access key as that can be used similarly to "bus" in routing.

### DIFF
--- a/obf_creation/rendering_types.xml
+++ b/obf_creation/rendering_types.xml
@@ -2478,7 +2478,6 @@
 		<routing_type tag="network" value="ncn" mode="register" base="true" relation="true"/>
 		<routing_type tag="network" value="icn" mode="register" base="true" relation="true"/>
 
-		<routing_type tag="access" mode="amend"/>
 		<!-- tag used as start with -->
 		<routing_type tag="access" mode="amend"/>
 		<routing_type tag="agricultural" mode="amend"/>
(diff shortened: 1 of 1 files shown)


vshcherb:
Indeed toll_booth is this . http://wiki.openstreetmap.org/wiki/Tag:barrier%3Dtoll_booth
The idea to have it with base routing in order to let all avoid paid roads options work

aceman444:
Shouldn't it then be <routing_type tag="barrier"> ? But before a recent patch of me to routing.xml (https://github.com/osmandapp/OsmAnd-resources/commit/8930f059be033470eba9b1000aa4800fcc62090c), barrier=toll_booth was NOT being evaluated for paid roads. So I do not understand what you say. Maybe it was an old leftover from times when toll_booths and paid roads were handled differently?

aceman444:
Eh, the editor did hide the tag in my first sentence:) Shouldn't it then be [routing_type tag="barrier"]? Or does the "tag" attribute cause it to also look in tag values, which is different to other rules in that file?

aceman444:
Hi, can I get the explanation so we could close this? :)

vshcherb:
Not sure that I get what needs to be explained

aceman444:
I proposed to remove the '<routing_type tag="toll_booth" mode="amend" base="true"/>' line but you said it may be needed to support "barrier=toll_booth" in routing.xml. But you still merged the patch thus removed the line. So I do not understand what you mean and how would toll_booth=* relate to barrier=toll_booth .

vshcherb:
Is there a tag toll_booth ? I didn't find documentation

aceman444:
I also this there is NOT a tag toll_booth. Therefore I removed it from the file. But you wrote as if that line was useful for something.

vshcherb:
No toll_booth here https://github.com/osmandapp/OsmAnd-resources/blob/master/obf_creation/rendering_types.xml

aceman444:
Well, you seem to have forgotten what was at the start of the discussion:) So if you have no further issues, I am fine with the current state of the file and we can close this. Thanks.

PR #320, opened 2015-05-16
add bicycle_repair_station

midnightcomm:
Bicycle repair station is an amenity for self service repair/fixing of bicycle.


xmd5a2:
specialphrases is obsolete and not used anymore. If you want to add a tag you should edit rendering_types.xml, poi_types.xml (resources repo) AND english phrases.xml (OsmAnd repo).

PR #324, opened 2015-05-23
Add support for Ukrainian language in voice guidance

jafd:
This pull request adds support for Ukrainian in voice guidance. It only contains live-recorded prompts, because no TTS engine whatsoever exists for this language, not an accessible one at the very least.

There are a few improvements underway (like improved numbers handling — I see quite some room for it), but the current config handles everything a recorded voice guidance would be expected to handle (that is, no street names and the like... But I think road names could be incorporated given some sweatwork).


vshcherb:
Thanks Yaroslav for hard work, though it would be very important to know the source of recorded voice. We need to be careful about licensing and maintenance issues.

jafd:
It's my own voice here, so the license is exactly as for the rest of the pull request (MIT).

xmd5a2:
I can not make it work. "Голосовые данные не поддерживаются текущей версией приложения".

jafd:
What version is it and logcat, please

jafd:
Ah, I think I know what it is.

OsmAnd reads in a file named "_config.p". In general case, gen_all.sh generates it from config.p after processing it, that is, feeding it into a TTS engine.

In this particular case, using a TTS engine is not possible, nor is it desirable. Just rename config.p into _config.p and you're good to go.

vshcherb:
Hello, before we generate new download file, it will be also good to know are you going to maintain this file and how much efforts will take from you. Usually we change 4-5 times a year, so we expect that will be possible to update in future, cause we can't record the same voice. In that sense it would be good to know how can we reach you.

xmd5a2:
Also, please redo folder structure. Our scripts does not support tree-like paths like voice/alerts, voice/arrivals. See other voices in OsmAnd-resources.

jafd:
Maybe it's the scripts that should be redone? There's really a lot of recorded files, and more to come (so that the prompts are not annoyingly unnatural – you have different intonation in the sentence's middle and in its end; I'm currently testing a new version which also tries to get rid of gaps while pronouncing numbers). This is a known plague for anyone who tries to make a decent domain-specific TTS system.

I might make a separate repository in which recordings are properly organized, and some sort of Makefile that will flatten them before integration to the main tree. But having to somehow mangle paths inside prolog files seems like a royal pain.

I can be contacted via yaroslav@fedevych.name. And yes, I'm committed to this project at least until we have a decent TTS engine with Ukrainian support (well, I might just signed a lifetime verdict).

jafd:
Also, how do you generate a download file? I really want to be able to test the whole workflow before I do a  pull request. Maybe the scripts are not this broken, after all...

vshcherb:
It is better to keep flat sctructure for consistency (./gen_voice.sh uk Ukranian ispeech uk)

jafd:
Also, a funny thing I discovered: there are noticeable gaps between media files being played, even if the files themselves are nearly gapless when their waveforms are simply concatenated (looks like the player just opens file after file every time, and this takes noticeable time). Should I make an issue in OsmAnd about this?

vshcherb:
If you know how to fix it , yes create a pull-request. Otherwise better not, it is hardly noticeable on recent devices

jafd:
I notice it on both Galaxy Note 3 (not quite an "old" device) and Nexus 7 from 2012. Just for the record.

Anyway, I'll better concentrate on the next voices version.

PR #325, opened 2015-05-24
In routing, disallow lift_gate and swing_gate access for cars by defa…

aceman444:
…ult, but also support destination and customers access.

According to http://wiki.openstreetmap.org/wiki/Barriers barriers with undocumented access default to access=no. So if there is no explicit access tag on the node, consider lift_gate and swing_gate as impassable with a car too. Lift_gate wiki even mentions cars cannot pass by default.

I expect some paid/mall parking areas to be blocked by a lift_gate tagged with access=customers or access=destination. So let's allow passing those by car, of course with a very low priority, so that the road is only chosen if it is the "last mile" to the destination.

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -385,6 +385,9 @@
 			<select value="100" t="barrier" v="height_restrictor"/>
 			<select value="100" t="barrier" v="sally_port"/>
 			<select value="1"   t="barrier" v="toll_booth"/>
+			<!-- These 2 shouldn't be allowed per the wiki (access=no is assumed) but people often do not add the proper access tags. -->
+			<select value="1000" t="barrier" v="lift_gate"/>
+			<select value="1000" t="barrier" v="swing_gate"/>
 			<!-- Without explicit access marking, barriers other than the listed values above are impassable to a car. -->
 			<select value="-1" t="barrier"/>
 
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -101,6 +101,8 @@
 			<select value="1"  t="motorcar" v="yes"/>
 			<select value="1"  t="motorcar" v="permissive"/>
 			<select value="1"  t="motorcar" v="designated"/>
+			<select value="1"  t="motorcar" v="destination"/>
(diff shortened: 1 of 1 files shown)


routing/routing.xml
```diff
@@ -349,21 +367,23 @@
...
 			<select value="1"  t="access" v="permissive"/>
+			<select value="1"  t="access" v="destination"/>
+			<select value="1"  t="access" v="customers"/>
 
 			<select value="5"   t="barrier" v="cattle_grid"/>
 			<select value="300" t="barrier" v="border_control"/>
 			<select value="300" t="barrier" v="bump_gate"/>
 			<select value="1"   t="barrier" v="entrance"/>
 			<select value="100" t="barrier" v="height_restrictor"/>
-			<select value="300" t="barrier" v="lift_gate"/>
```
vshcherb: This is too strict limitation, I know some areas where lift_gate is the only possible way to enter and of course you should have permission, but people who put target there, certainly have that permission

aceman444:
Yes, then there should be access=destination (or private) on the barrier node. It is probably the same problem with 'gate'. People do not put any access tag on it and some expect it to be passable, some not.

vshcherb:
Definitely they won't put, but the problems causing this missing tag are quite big, so I wouldn't go the path disallowing lift_gate

aceman444:
I have added new change with a question.

aceman444:
So what if we made the value be "1000000" to really discourage OsmAnd to use the lift gates?

vshcherb:
We shouldn't discourage that's the whole point. I think event to revert to 300, values > 1000 makes the whole routing slower (and even not linearly).

aceman444:
Why not discourage? The unmarked gates should only be used if there is no other way. I do not mean prevent (as in value=-1), just discourage.

vshcherb:
That's not possible discourage, you could put a high price for point, but that slows down the routing search. So unless we don't see concrete example where penalty > 300 makes routing better, we should not use it.

PR #328, opened 2015-05-31
Prioritize foot=official paths in pedestrian mode

aceman444:
Foot=official is used on officially marked paths that pedestrians should (and in some countries even must) use. So give it some priority. Also remove rules detecting "pedestrian=no" as that is an outdated tag no longer used much. We can support it via tag rewriting in rendering_types.xml.

### DIFF
--- a/obf_creation/rendering_types.xml
+++ b/obf_creation/rendering_types.xml
@@ -2816,6 +2816,7 @@
 		<routing_type tag="enforcement" mode="amend"/>
 		<routing_type tag="ferry" mode="amend"/>
 		<routing_type tag="foot" mode="amend"/>
+		<routing_type tag="pedestrian" tag2="foot" mode="replace"/>
 		<routing_type tag="goods" mode="amend"/>
 		<routing_type tag="hgv" mode="amend"/>
 		<routing_type tag="horse" mode="amend"/>
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -423,9 +423,9 @@
 			<select value="-1" t="route" v="ferry">
 				<if param="avoid_ferries"/>
 			</select>
-			<!-- <select value="-1" t="access" v="no"/> -->
 
 			<select value="-1" t="bicycle" v="no"/>
+			<select value="-1" t="bicycle" v="use_sidepath"/>
 			<select value="1"  t="bicycle" v="yes"/>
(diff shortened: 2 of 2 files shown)


vshcherb:
Removing tags will be quite operation and may be not needed.
Also prioritizing roads by 1.5 doesn't seem logical, cause for foot navigation people are not only interested in official routes

aceman444:
I do not remove support for pedestrian=*, even though I would vote for it, as it is used only 600 times in the world and is undocumented, so there is no need to support it. I just move it to the obf creation. In that way routing should get faster.

Normal footways get priority=1, official ones just get 1.5. We can argue about the exact value.
But what is wrong with the proposal? It does not mean OsmAnd will only use official paths.

As I said, in some countries using foot=official is mandatory and you must not use another parallel way. So how do you want to solve this without at least increasing the prority?

vshcherb:
I'm not very familiar with foot=official to be honest, but to be honest I don't like we change parameter with a magic guess. If it is official, let's download and try it in these countries and just post an example screenshot here. All coefficients > 1 are not correct in sense of A* and to be honest they will be cut inside if the speed goes over maximum. 

To test the routes please use OsmAndMapCreator which works on every platform, so I think we need to have more testing around it, so that's may be the main reason to not accept this pull-request straight-forward. I will keep it open, so we could get back to testing routines. 

Thanks,
Victor

aceman444:
OK, can you merge in the "pedestrian" part of the patch?

vshcherb:
It is better to wait for complete solution.

aceman444:
The 2 changes are not related. So I can split it out into separate pull request.

aceman444:
The patches 183ea6c and f7f8e47 should just be deleted here and the patch 1811840 is to be applied. Then it should have no conflicts.

aceman444:
Also see http://wiki.openstreetmap.org/wiki/Proposed_features/use_sidepath#Routing where OsmAnd is specifically mentioned as not supporting =use_sidepath yet.

aceman444:
I do not see which patch got merged. Usually there is a status message in the pull request. But I do not see it here and the file is unchanged at master branch.

aceman444:
It looks to me all parts of this got reworked and merged in other pull requests. The last use_sidepath part is in PR #343. So this can stay closed, thanks.

PR #332, opened 2015-06-12
Improve routing on hiking trails

n76:
See discussion at https://groups.google.com/forum/#!topic/osmand/An1wQpP7mZ4

This change allows user to avoid other highways besides motorways and to give a higher priority to paths and footways.


aceman444:
Have you tried to just set priority to 1.1 for foot=yes before the [select value="1"/] rule?

n76:
With respect to the [select value="-1" t="pedestrian" v="no"/]  rule, that was in the file before. You are right that it probably ought to be removed. But there are lots of rules in there that make no sense to me, for example there is in another area a reference to highway=services (not documented in OSM) in addition to highway=service. I was attempting to make my changes as small as possible and not get involved in a wholesale clean up of the entire file.

With respect to your comment "Have you tried to just set priority to 1.1 for foot=yes before the [select value="1"/] rule?", I am not sure which line in the file you are referring to.

aceman444:
But you neverthless cleaned up the whole file, or your editor did, removing trailing whitespace in unrelated lines.

I do not understand why you add the pedestrian line, just because it "was in the file before". Please work on the current state of the file in the master branch, not some version from OsmAnd v2.0.x. Yeah, the highway=services line is strange, but better not touch it if we do not understand it. services usually are areas so routing over them may happen on their outline, similar to highway=pedestrian areas.

I meant to add 
			[select value="1.1" t="foot" v="yes"/]
			[select value="1"/]
in the priority block for pedestrian.

Also, I am not sure you should prioritize highway=path unconditionally (with value = 1.2). As it may be unpaved or really bad surface. We should add some prioritization based on surface tag (as done for car), but that is another can of worms :)

n76:
Okay, I think I understand your requests and have made those changes.

n76:
With respect to prioritizing paths based on surface/condition, it looks like both bicycle and auto routing simply have a checkbox for "avoid unpaved roads". In some regards hiking paths could be easier as there is a numeric sac_scale tag defined in OSM (I haven't checked to see if it gets into the osmand database). Some hikers would actually prefer the more difficult routes, so a single "avoid" checkbox might not fit but if it is possible to have a multiple select, either pull down or radio buttons) to specify the maximum sac scale that should be considered, I think there could be some value add. But, as you wrote, 
that is another can of worms.

aceman444:
The commit 19b9df8 does not do what I proposed. You put the foot=yes line above highway=primary which will have the opposite effect to what you want to achieve.

n76:
Sorry that I'm slow at this. :)

I am also beginning to think that the "prefer paths" that I added should not be there as the priority of highway=pedestrian, highway=path and highway=footway are all bumped when selecting pedestrian routing anyway.

So, if I move the foot=yes to below the highway priority lines and remove the logic for prefer paths, would that suit your requirements?

aceman444:
I would also separate the avoid_motorways (motorways ad trunk) and avoid_highways because some users may want to use normal roads, but not motorways (it is even forbidden in some countries). Also, I do not think you really should forbid normal roads with the setting as sometimes footways, sidewalks and similar are mapped as good as roads. Or there often are missing some connections of roads and footways. So OsmAnd may produce a very long artificial route if you simply forbid roads. I would suggest to just lower their priority if the settings is set, e.g. to 0.1 .
It all depends on preferences. Would even hikers in a national park like to avoid going 1km over a road if it saves them 10km over paths? Some would, some not.

n76:
Is there an explanation of what the priority value means? Is it something like a multiplier to put on the cost on using the way? That is if a highway=residential has a priority of 1 and a highway=path has a priority of 2 would the path be used as long as it is less than twice as long as the residential?

n76:
Okay, the version on this last push seems to do what I want: If avoid highways is selected route is over a longer hiking trail. If it is unselected route is over highway. And, if my guess on the meaning of priority is correct, then it might find a route that has a short section of highway.

vshcherb:
I think it has been a while we've been changing routing parameters with a magic guess, since last month I'm trying to make only reasonable changes.  For example, 
+<select value="1.2" t="highway" v="path"/>
+<select value="1.2" t="foot" v="yes"/>
for these 2 lines changes, please find a real case scenario (just screenshots with map creator), that changes situation to better route. It will really setup a good base for future tests and prevent changing parameters blidingly. 
Thanks.

n76:
"for these 2 lines changes, please find a real case scenario (just screenshots with map creator), that changes situation to better route." (not sure which 2 line changes) but here are two screen shots with 2.0.4 the with an updated routing.xml set for pedestrian mode. First on with "avoid highways" off:
screenshot_2015-06-15-05-50-36

The second one with "avoid highways" on:
screenshot_2015-06-15-05-51-34

Is this the type of use case you are referring to? (First time I've tried to attach images to a reply in GitHub, I hope they work.

vshcherb:
Yes this is good example, even though the pictures could be scaled down, they are too big ) Anyway why would you think that 5.5 miles route is better than 3.89 miles? What is the exact changing causing it, all together or that you lower all secondary/primary roads

n76:
Screen shots were direct from my phone. I guess it is a good thing I don't have a tablet. :)

I could post some pictures of what the trail looks like versus what the highway looks like, but when you are hiking you generally prefer to be on a nice scenic trail away from the automobile traffic. This change, by optionally lowering the highway priority if the user so desires, allows you to do that.

sonora:
Frankly, I think there will always be things you will have to decide
"subjectively" by looking at the map, or even by looking at the "real
world" (like you suggested "via pictutes" ... :-)  )

How should an algorithm determine how much of a detour you are willing to
take to avoid an e.g. tertiary road? 1 mile? 2 mile? 3 miles? It seems
rather arbitrary to have a parameter for this which would again be trial
and error, and extremely situation-dependent for every user to figure out.
So if you think you see a "nicer" alternative to take while the routing
algorithm suggested a shorter or faster route, just take it, and OsmAnd
will re-calculate eventually anyway to adjust for where you are going....
my 5 cents. - Hardy

n76:
So a change that simply adds an option, perhaps not totally optimized as yet, that makes Osmand much more hiker friendly in areas where there are highways is to be rejected because it will result in a route that is the most direct one that avoids (but does not prohibit) routing over highways?

If it were to replace existing pedestrian routing I could see the argument, but it requires having a checkbox specifically selected before any changes in routing behavior occur. And that checkbox clearly states "avoid highways" so the user should know what they are getting.

sonora:
Yes, avoid highways is different, and clear, in analogy to avoid motorways
(although I am afraid any routes will not exist then). I was under the
impression that you wanted to lower highway priority in an unclear fashion
where sometimes you would get routes with or without highway portions - and
that would be rather intransparent or unpredictable, I could imagine?

sonora:
PS: Quick sanity check here: Does anybody really use navigation for hiking?
I do a lot of hiking, and have all sorts of criteria to pick my way,
scenery, viewpoints, etc., none of which an algorithm would help me with
much ... :-)

vshcherb:
To be honest I'm fine with addiing option comparing to other profiles, pedestrian doesn't really have options, but the biggest changes is in <avoid> section. So I wonder, how & why these 2 lines affect the routing, if you delete them, will the route change? Are they done in follow analogy manner, but then I'm not sure that analogy was correct as well.

select value="1.2" t="highway" v="path"/>
<select value="1.2" t="foot" v="yes"/>

n76:
"Does anybody really use navigation for hiking?" I use it for showing the distance remaining to my destination. I guess I could come up with a bunch of waypoints to force it along the path I plan to take but for things like returning to the car at the trail head I usually go the shortest/fastest non-highway direction.

With respect to the addition of 'select value="1.2" t="highway" v="path"', it seemed that it ought to be treated much the same as highway=footway.

vshcherb:
>>  it seemed that it ought to be treated much the same as highway=footway.
This is exactly my point, we shouldn't change routing parameters based on our gut feeling. If it doesn't have any visible difference, it is better not to have it at all. 
P.S. I think even for footway 1.2 is not needed

vshcherb:
This pull request seems to be very old, it make sense to recreate it if it still makes sense.

PR #334, opened 2015-06-24
Also detect ATMs tagged as "atm=yes"

jrudolph:
Many ATMs in buildings are only tagged as "atm=yes".

See http://wiki.openstreetmap.org/wiki/Tag:atm%3Dyes

Not sure if this is the best way to do it as I'm working on this file for the first time.

### DIFF
--- a/poi/poi_types.xml
+++ b/poi/poi_types.xml
@@ -819,7 +819,10 @@
 		</poi_category>
 		<poi_category name="finance" default_tag="finance">
 			<poi_type name="bank" tag="amenity" value="bank"/>
-			<poi_type name="atm" tag="amenity" value="atm"/>
+			<poi_filter name="atm">
+				<poi_type name="atm_standalone" tag="amenity" value="atm"/>
+				<poi_type name="atm_at_building" tag="atm" value="yes"/>
+			</poi_filter>
 			<poi_type name="payment_terminal" tag="amenity" value="payment_terminal"/>
 			<poi_type name="money_lender" tag="shop" value="money_lender"/>
 			<poi_type name="bureau_de_change" tag="amenity" value="bureau_de_change"/>


poi/poi_types.xml
```diff
@@ -819,7 +819,10 @@
 		</poi_category>
 		<poi_category name="finance" default_tag="finance">
 			<poi_type name="bank" tag="amenity" value="bank"/>
-			<poi_type name="atm" tag="amenity" value="atm"/>
+			<poi_filter name="atm">
+				<poi_type name="atm_standalone" tag="amenity" value="atm"/>
+				<poi_type name="atm_at_building" tag="atm" value="yes"/>
```
xmd5a2: atm=yes used not only with buildings so I disagree with name http://overpass-turbo.eu/s/a5K

poi/poi_types.xml
```diff
@@ -819,7 +819,10 @@
 		</poi_category>
 		<poi_category name="finance" default_tag="finance">
 			<poi_type name="bank" tag="amenity" value="bank"/>
-			<poi_type name="atm" tag="amenity" value="atm"/>
+			<poi_filter name="atm">
```
xmd5a2: It will not work this way. It is better to keep tag/value here

xmd5a2:
You forgot to add atm=yes to rendering_types as additional tag. But additional tag can be only linked to particular main tag/value. So in theory you should add atm=yes as poi_additional (!) to all poi_types which can be used with atm=yes.
Also you need to update https://github.com/osmandapp/Osmand/blob/master/OsmAnd/res/values/phrases.xml

poi/poi_types.xml
xmd5a2: Also change poi_filter to poi_type

jrudolph: This was just a shot in the dark. Is there an explanation for `poi_filter` and the other tags? 

Is it because one object can only be matched once either by `poi_type` or `poi_filter`? And if you need it to be matched several times you need to use `poi_additional`?

xmd5a2:
Merged manually another way https://github.com/osmandapp/OsmAnd-resources/commit/7a943b47b1426e803e1f6a91c33ad0ab344cc671

jrudolph:
Will your committed change also show `atm=yes` objects in the POI search for ATMs? That's what I'm actually more interested in (and I guess most people because it seems the majority of all ATMs are mapped as `atm=yes` objects of banks).

poi/poi_types.xml
xmd5a2: poi_filter used for predefined filters which are shown in search UI. poi_type - regular poi. poi_additional it "addition" to main tag. It is because our map data system.

jrudolph:
Also with that change won't a lot of shop and bank entities be now rendered as ATMs? As said before, IMO the rendering is less important than being able to find ATMs tagged as `atm=yes`. (Thanks btw for considering this at all and the speedy response)

xmd5a2:
In future yes. There is an issue with entity_convert and POIs which we are planning to resolve.

xmd5a2:
I'll test it, but it is possible to lower icon and text priority.

xmd5a2:
Yes, it causes problems with old versions so I reverted this commit.

xmd5a2:
This change needs to be done another way which is not implemented yet.

PR #335, opened 2015-06-26
also regard ATMs at banks and post office mapped as `atm=yes` when searching for ATMs

jrudolph:
Follow up to #334. 

Do you think this one would work at least till a better solution would be implemented? If not, I'll patiently wait for a better implementation, I promise ;)

[Currently, 92 % of all `atm=yes` usages are in combination with `amenity=bank`](http://taginfo.openstreetmap.org/tags/atm=yes#combinations).

Thanks for considering this :)

### DIFF
--- a/poi/phrases/de/phrases.xml
+++ b/poi/phrases/de/phrases.xml
@@ -36,6 +36,9 @@
 	<string name="poi_astronomical_observatory">Observatorium</string>
 	<string name="poi_athletics">Sport</string>
 	<string name="poi_atm">Geldautomat</string>
+	<string name="poi_atm:standalone">Geldautomat (frei stehend)</string>
+	<string name="poi_atm:at_bank">Geldautomat (in Bank)</string>
+	<string name="poi_atm:at_post_office">Geldautomat (in Post)</string>
 	<string name="poi_attraction">Attraktion</string>
 	<string name="poi_atv">Quad-Store</string>
 	<string name="poi_audiologist">Gehörspezialist</string>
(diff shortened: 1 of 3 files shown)


xmd5a2:
There is no smart way to do this for now. Let's wait for improving entity_convert functionality.

vshcherb:
To be honest that was a case for a long time and from my experience in 100% when I was searching for ATM and went to the same bank, I saw an atm there. So it is kind of known pattern, I think implementing this logic could even confuse old users

jrudolph:
@xmd5a2 Ok, so this change wouldn't work at all?

@vshcherb I agree, it's a known pattern. So, do you suggest the right way to look for an ATM in osmand is to search for a bank, for a post office or a standalone ATM separately? Is that explained somewhere? It doesn't seem to be the most straight-forward way IMO...

aceman444:
Not all the banks have an ATM. On the other hand, you probably can draw money from that bank so you get your cash as was the whole point of searching for the ATM. Unless there are services in the ATM that the bank won't do (like check your account balance free of charge or top-up prepaid cell phone plan).

I vote for atm=yes being supported when searching for ATMs.

xmd5a2:
https://github.com/jrudolph No, whis wouldn't work because you have not modified rendering_types.xml which describew what data to extract from OSM database. But don't do whit. I agree with 
https://github.com/aceman444 that atm=yes should be supported when searching for ATMs. But for now we have no way to do this without changing our maps processing code. These changes are planned.

PR #337, opened 2015-07-13
Do not avoid highway=cycleway with conflicting vehicle tag

ssiegel:
For example http://www.openstreetmap.org/way/126308212 is tagged
highway=cycleway and vehicle=agricultural, so this part of the national
cycle network was excluded from routing.

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -435,7 +435,10 @@
 			<select value="1"  t="bicycle" v="destination"/>
 			<select value="1"  t="bicycle" v="official"/>
 
-			<select value="1" t="highway" v="cycleway"/>
+			<select value="-1" t="highway" v="motorway"/>
+			<select value="-1" t="highway" v="motorway_link"/>
+			<select value="1"  t="highway" v="cycleway"/>
+			<select value="1"  t="highway" v="path"/>
 
 			<select value="-1" t="vehicle" v="no"/>
 			<select value="-1" t="vehicle" v="agricultural"/>
@@ -451,8 +454,6 @@
 			<select value="1"  t="access" v="yes"/>
 			<select value="1"  t="access" v="permissive"/>
 
-			<select value="1" t="highway" v="motorway"/>
-			<select value="1" t="highway" v="motorway_link"/>
 			<select value="1" t="highway" v="trunk"/>
 			<select value="1" t="highway" v="trunk_link"/>
 			<select value="1" t="highway" v="primary"/>
@@ -466,7 +467,6 @@
 			<select value="1" t="highway" v="unclassified"/>
 			<select value="1" t="highway" v="service"/>
 			<select value="1" t="highway" v="track"/>
-			<select value="1" t="highway" v="path"/>
 			<select value="1" t="highway" v="living_street"/>
 			<select value="1" t="highway" v="pedestrian"/>
 			<select value="1" t="highway" v="footway"/>
@@ -644,12 +644,16 @@
 			<select value="1"  t="foot" v="destination"/>
 			<select value="1"  t="foot" v="official"/>
 
+			<select value="-1" t="highway" v="motorway"/>
+			<select value="-1" t="highway" v="motorway_link"/>
+			<select value="1"  t="highway" v="path"/>
+			<select value="1"  t="highway" v="pedestrian"/>
+			<select value="1"  t="highway" v="footway"/>
+
 			<select value="-1" t="access" v="no"/>
 			<select value="1"  t="access" v="yes"/>
 			<select value="1"  t="access" v="permissive"/>
 
-			<select value="1" t="highway" v="motorway"/>
-			<select value="1" t="highway" v="motorway_link"/>
 			<select value="1" t="highway" v="trunk"/>
 			<select value="1" t="highway" v="trunk_link"/>
 			<select value="1" t="highway" v="primary"/>
@@ -664,10 +668,7 @@
 			<select value="1" t="highway" v="unclassified"/>
 			<select value="1" t="highway" v="service"/>
 			<select value="1" t="highway" v="track"/>
-			<select value="1" t="highway" v="path"/>
 			<select value="1" t="highway" v="living_street"/>
-			<select value="1" t="highway" v="pedestrian"/>
-			<select value="1" t="highway" v="footway"/>
 			<select value="1" t="highway" v="byway"/>
 			<select value="1" t="highway" v="platform"/>
 			<select value="1" t="railway" v="platform"/>
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -435,6 +435,8 @@
 			<select value="1"  t="bicycle" v="destination"/>
 			<select value="1"  t="bicycle" v="official"/>
 
+			<select value="1" t="highway" v="cycleway"/>
+
 			<select value="-1" t="vehicle" v="no"/>
 			<select value="-1" t="vehicle" v="agricultural"/>
 			<select value="-1" t="vehicle" v="forestry"/>
@@ -461,7 +463,6 @@
 			<select value="1" t="highway" v="tertiary_link"/>
 			<select value="1" t="highway" v="road"/>
 			<select value="1" t="highway" v="residential"/>
-			<select value="1" t="highway" v="cycleway"/>
 			<select value="1" t="highway" v="unclassified"/>
 			<select value="1" t="highway" v="service"/>
 			<select value="1" t="highway" v="track"/>


aceman444:
What is wrong in avoiding that road? There is no access for normal bikers. OsmAnd must not route there. If you know the situation on that road better, please fix the tags (e.g. add bicycle=yes). But the change to OsmAnd routing seems wrong to me.

ssiegel:
http://wiki.openstreetmap.org/wiki/Tag:highway%3Dcycleway says:

> Tagging a way with highway=cycleway implies that the route is designated for bicycles.

and

> Implies
> bicycle=designated

Both bike routing engines on openstreetmap.org (GraphHopper and MapQuest) are happy to use (and prefer) this way.

Many ways affected, see http://overpass-turbo.eu/s/aps. Judging from context, most if not all of them are accessible to bikes. To me it seems unfeasible to change them all, especially since OSM data should not target specific renderers or routers.

aceman444:
Well, in my understanding "implies" only applies if there is no explicit tagging of the access. I would think in this case explicit tagging overrides any implicit assumptions.

Thanks for the overpass-turbo query. To me it seems there are only 255 (not "many") in whole of Europe, in some countries only sporadic uses. To me that seems more like tagging bugs or that those cycleways are really currently closed/inaccessible.

Can you point to a discussion on some mailing list that the access rules should be interpreted in the way you want it? (I.e. adding bicycle=designated to the explicit tags and then evaluating the result).

ssiegel:
I cannot find explicit discussion for this specific case. Most of the discussion refers to the wiki and interprets highway=cycleway as shorthand form, e.g. this thread: https://lists.openstreetmap.org/pipermail/tagging/2010-January/001062.html

There is additional information here:
http://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access-Restrictions#Default shows "designated" in non-italic in row "cycleway", column "bicycle", meaning that the value is not inherited from generic access tags.

ssiegel:
Oh, I see the wiki table has non-italic "Yes" for row "path", which means the line for "highway=path" should be moved up along with the line for "highway=cycleway", and maybe "highway=track" (proposal for Germany) as well. Should I update my pull request?

aceman444:
So you want to avoid motorway even if for some reason it would have bicycle=yes ? OsmAnd is currently designed for the whole world and does not have country specific rules. You can create your own routing.xml file and use it in OsmAnd with your special rules. But you can't modify the official routing.xml just because 10 countries have some special rule.

I simply don't agree with the whole idea of the patch. If we went by your logic, then highway=footway is designated for pedestrians. And highway=primary is designated for motorcars. So now what if you have highway=footway+access=no ? And highway=primary+access=no? So then in your evaluation, the footway is still accessible to foot (as an added foot=designated would override access=no) and the primary road is accessible to motorcars. That is simply not true (both combinations are used in OSM data). I'd really like to see anybody that shares your evaluation method.

ssiegel:
As I understand it, a motorway having bicycle=yes is not ignored, because bicycle=yes is checked first. Perhaps I missed something, in which case I ask you to show me my mistake.

The ten countries I enumerated are listed on the linked wiki pake as having similar rules for highway=track, but as you can see when looking at my proposed patch, I did not touch any rules for highway=track.

highway=footway+access=no will be considered for pedestrian routing according to the rules on the wiki page, and it will be considered when using my patch. It should be tagged with foot=no if it is to be ignored for pedestrian routing.

highway=primary+access=no won't be considered for routing, and my patch doesn't touch any highway=primary routing decisions.

What my patch does change is stop routing bicycles or pedestrians over ways tagged highway=motorway, which is correct in my view.

Maybe my first version of the patch was incomplete, but it was correct according to

- the OSM wiki
- at least two other routing engines and
- personal experience in specific example
- the German “Verkehrszeichen Tool”, see http://osmtools.de/traffic_signs/?signs=244,1026-36. The shown combination (same as in my example) allows bicycles according to StVO, does not allow bicycles using Osmand according to the generated tags, but does so using my patch.

I think I gave enough evidence in this case. If you disagree, I ask you to provide some evidence/wiki entries/discussion threads to support your view. Are you able to show me a routing engine (besides Osmand) that avoids way 126308212 for bicycle routing? 

My second commit here just extends on this and tries to make the change more consistent and complete.

aceman444:
I think I now see where you get your proposed evaluation.
It can be seen at http://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access-Restrictions and http://wiki.openstreetmap.org/wiki/Computing_access_restrictions .

Well, neither of them is agreed upon yet, the latter one marked as outdated and for deletion (while the referred "conditional access rules" seem not to contain your proposed model).

I have always tagged according to the model described at http://wiki.openstreetmap.org/wiki/Talk:Computing_access_restrictions#Alternative_evaluation_model which is much easier to grasp to a human. I have not seen that comment until today, but I have inferred it from reading http://wiki.openstreetmap.org/wiki/Key:access . There are not enough examples there to support your or mine evaluation model.

So can you first get those wikipages with your evaluation model get voted upon and approved?

ssiegel:
The wiki says “Default is a proposal”, which means there is currently no globally unique system of default access restrictions. There are, as I understand it, different systems in different countries currently in use, some of them more or less officially decided upon in national OSM communities, some not. Therefore it is currently impossible to have one single system giving correct routing decisions all of the time.

I understand the “default” table as an attempt to find a system that works “most of the time” and might be a candidate for a future global default, therefore something one can use right now in case there is need for an approximation of a unique default access restriction system. I don't see why any other (equally little officially agreed upon) method might be better suited for that use-case than this table.

This is not “my” evaluation model, and I find it ridiculous to ask me to get anything voted for and approved. I am just a more or less ordinary user who is annoyed by Osmand making suboptimal bike routing decisions, last time the last weekend on the mentioned way. I tracked down the root of the problem, I fixed it, and I wanted to share the result of my work with you so other users have the chance to be less annoyed by the router in the future.

First I had to prove my point in a not very constructive discussion (“fix the tags” while it works with the other available routers), now I need to get my suggestion “voted upon and approved”. I will just stop here and continue to use my own local version of routing.xml. Thank you.

ssiegel:
For reference, this is how GraphHopper solves this: https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java#L237

vshcherb:
I don't think the argument that it works in some routing engines is good one. It used to work in OsmAnd as well, but that doesn't mean it is correct, sometimes we need to change the map if it is not correct. 
1) I read about agricultural/forestry and I mostly think these are restrictions for motor vehicles (it even says in osm wiki), I'm not 100% sure, it is better to check on the ground. I disabled these checks for now.
2) I don't agree on disabling motorway roads, we have a special parameter here. Sometimes it is an only way and sometimes motorway has special sideway. Just let's not project motorways we know on all motorways in the world. I definitely know couple of motorway where people walk on the sideway.

aceman444:
>ssiegel commented 22 hours ago
>The wiki says “Default is a proposal”, which means there is currently no globally unique system >of default access restrictions. There are, as I understand it, different systems in different >countries currently in use, some of them more or less officially decided upon in national OSM >communities, some not. Therefore it is currently impossible to have one single system giving >correct routing decisions all of the time.
Exactly, so why do you think your system is better and want to change OsmAnd to it?

>I understand the “default” table as an attempt to find a system that works “most of the time” and >might be a candidate for a future global default, therefore something one can use right now in >case there is need for an approximation of a unique default access restriction system. I don't >see why any other (equally little officially agreed upon) method might be better suited for that >use-case than this table.
I find that system very fragile to new users. If there is highway=footway+access=no and somebody demotes it to highway=path, suddenly it becomes accessible to bikes and horses? Hardly anybody will remember to add bicycle and horse=no. I expect this would just produce a ton of wrong accesses where not intended. I have never seen tagging according to the "defaults" system in practice. Never something like highway=footway+access=no+foot=no. Do you have an example city where it is done the way you propose?

But I agree the system OsmAnd currently uses may be "equally little agreed upon". But then why switch from one unofficial system to another?

>This is not “my” evaluation model, and I find it ridiculous to ask me to get anything voted for and >approved. I am just a more or less ordinary user who is annoyed by Osmand making suboptimal >bike routing decisions, last time the last weekend on the mentioned way. I tracked down the root >of the problem, I fixed it, and I wanted to share the result of my work with you so other users >have the chance to be less annoyed by the router in the future.
How can you determine something is "suboptimal" and that you "fixed it" when there is no agreed upon system how to evaluate the rules? I then find it ridiculous that you ask all the others to have a changed app to use an unapproved evaluation system that some (probably a minority) users thought up.

>First I had to prove my point in a not very constructive discussion (“fix the tags” while it works >with the other available routers), now I need to get my suggestion “voted upon and approved”. I >will just stop here and continue to use my own local version of routing.xml. Thank you.
Sure, many people do that, as OsmAnd does not contain country specific rules and exceptions. I also find that a pity and would vote for it where possible.

ssiegel:
I think I said enough about this specific issue. This is not about “my way” or “your way”, this is about improving Osmand. If you just show me one example where my commit 26a9c06 causes clearly worse bike routing, then maybe I am willing to talk about it again.

I am simply surprised about the tone and style of discussion in the Android community – on XDA, Xposed, and now here. In the Linux and KDE communities where I participated in the past, I was used to a friendly atmosphere welcoming help and suggestions from anybody who wants to give something back, open to constructive discussion about any proposal. But it's probably time for me to move on.

ssiegel:
For highway=footway+access=no+foot=no just use overpass turbo again: http://overpass-turbo.eu/s/asc

vshcherb:
I think highway=cycleway and access=no, should always be forbidden. It is very simple and clear, by access=private or access=no, it means that route is special, it could be dangerous or military zone. As a mapper I would expect it. I found these examples next to me http://www.openstreetmap.org/way/34996986, http://www.openstreetmap.org/way/39523163 and your commit allows to use them, which I think is wrong, but I need to check these roads myself..

PR #338, opened 2015-07-17
Discourage ways with *=use_sidepath

aceman444:
See http://wiki.openstreetmap.org/wiki/Tag:bicycle%3Duse_sidepath for how the tag is to be used.
At http://wiki.openstreetmap.org/wiki/Proposed_features/use_sidepath#Routing they specifically call out OsmAnd as not supporting the tag yet.
We could use value="-1" for some countries, while for some countries using such ways for foot/bicycle may still be allowed. So reducing the priority may be the middle ground.

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -476,6 +476,7 @@
 		</way>
 
 		<way attribute="priority">
+			<select value="0.1" t="bicycle" v="use_sidepath"/>
 			<if param="avoid_unpaved">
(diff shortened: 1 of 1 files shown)


vshcherb:
Please recreate pull-request or redo. 
>> We can’t automatically merge this pull request.

aceman444:
I don't know how to update a patch via the github webinterface so I had to create a new pull request #343 .

PR #340, opened 2015-07-21
Allow shuttle trains

ionum:
Create new option to avoid shuttle trains (similar to ferries)

Resolves https://github.com/osmandapp/Osmand/issues/1308

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -62,6 +62,7 @@
 		<parameter id="avoid_toll" name="Avoid toll roads" description="Avoid toll roads" type="boolean"/>
 		<parameter id="avoid_unpaved" name="Avoid unpaved roads" description="Avoid unpaved roads" type="boolean"/>
 		<parameter id="avoid_ferries" name="Avoid ferries" description="Avoid ferries" type="boolean"/>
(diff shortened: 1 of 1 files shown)


aceman444:
There are only 18 uses of shuttle_train in the whole OSM world. Also, there is no wiki page about how to use it. How do you know the routing over it actually works correctly?

ionum:
I tested it. It generally works. There may be only 18 shuttle_trains, but they are important ways in their regions. OSRM and other routing engines already use it.

If you want to test it:

From the mainland to the island of Sylt, connected by the Hindenburgdamm, only by train and foot (Germany, Schleswig-Holstein)
From Kandersteg to Brig (Switzerland)
From Brig to Domodossola (Switzerland & Italy Piemont)

BTW: Option "Avoid shuttle_train" has to be translated. I can deliver german & french. How can I help

aceman444:
So if the trains are important, can you create the wiki page so that everybody knows how to tag them. Then everybody can also see why OsmAnd is doing what it is doing now with them :)

ionum:
See http://wiki.openstreetmap.org/wiki/Proposed_features/shuttle_train

aceman444:
Nice, thanks. Maybe you can now add OsmAnd to the list on that page :)

ionum:
Done

PR #341, opened 2015-08-03
Update phrases.xml

dktzde:
corrected German: Fahrgemeinschaft -> Carsharing Standort


xmd5a2:
There is an easier way to translate. Use https://hosted.weblate.org/projects/osmand/

xmd5a2:
Besides, this is not original file, just a copy for development purposes

dktzde:
Thanx a lot - but I cannot find the string on weblate. So I will give up now to correct the misleading translation. (maybe you can help?)
see also: https://github.com/osmandapp/Osmand/issues/1517

xmd5a2:
See https://github.com/osmandapp/Osmand/blob/master/OsmAnd/res/values-de/phrases.xml
This is correct and actual file.

dktzde:
Thanx!

PR #343, opened 2015-08-08
Discourage ways with *=use_sidepath (copy of issue #338)

aceman444:
See http://wiki.openstreetmap.org/wiki/Tag:bicycle%3Duse_sidepath for how the tag is to be used.
At http://wiki.openstreetmap.org/wiki/Proposed_features/use_sidepath#Routing they specifically call out OsmAnd as not supporting the tag yet.
We could use value="-1" for some countries, while for some countries using such ways for foot/bicycle may still be allowed. So reducing the priority may be the middle ground.

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -433,7 +433,6 @@
 			<select value="-1" t="route" v="ferry">
 				<if param="avoid_ferries"/>
 			</select>
-			<!-- <select value="-1" t="access" v="no"/> -->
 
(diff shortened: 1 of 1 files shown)


vshcherb:
Didn't understand why they are so deprioritized

vshcherb:
Got it

aceman444:
The way on which this tag is put should not be used as there should be an official way for the transport very close to it. So by lowering the priority OsmAnd routing hopefully picks up the way with normal priority.

PR #344, opened 2015-08-21
Update routing.xml with better highway transitions

PinkDuck:
Currently has questionable time penalties causing very unhelpful fastest routing outcomes:

35 seconds from trunk_link to primary/primary_link?
50 seconds from primary/primary_link to secondary/secondary_link?
50 seconds from secondary/secondary_link to tertiary/tertiary_link?
50 seconds from tertiary/tertiary_link to anything lower? e.g. unclassified/residential.

The time differences should diminish as the road prominence reduces as they are non-linear in capacity.

Avoiding the 50 second penalty from tertiary to unclassified, and the others, will restore sane routing while keeping the intended desire of this feature to simulate delays leaving higher load roads.

It would still be better to consider speed limit and number of lanes to derive a penalty than the classification types.

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -196,13 +196,13 @@
 			<select value="10" t="highway" v="motorway_link"/>
 			<select value="10" t="highway" v="trunk"/>
 			<select value="15" t="highway" v="trunk_link"/>
-			<select value="50" t="highway" v="primary"/>
-			<select value="50" t="highway" v="primary_link"/>
-			<select value="100" t="highway" v="secondary"/>
-			<select value="100" t="highway" v="secondary_link"/>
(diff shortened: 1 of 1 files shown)


vshcherb:
This issue requires proper testing, we have 2 open bugs which certainly causing troubles and not related exactly to these numbers. Removing these penalties will fix the issue, but also makes routing slower and also affects some routes fixed by this feature. 

I will avoid merging pull-request until we figure out how to test and maintain tests, unfortunately it is not done yet.

PinkDuck:
I am pretty certain that a 50 second time penalty from tertiary to unclassified is causing troubles and not doing much to fix the other routes. Note I am not proposing their removal, just more sensible values. Hopefully the fastest car routing problems will be addressed soon.

PR #345, opened 2015-08-24
Support smoothness tag in rounting.xml

aceman444:
I propose to add support for smoothness key. According to http://wiki.openstreetmap.org/wiki/Key:smoothness values of impassable down to horrible are not passable to normal car. I think that is what the "car" profile in OsmAnd presents. The value of very_bad for "high clearance car" is questionable. So I allow it for users that do not care about unpaved roads. Those probably have something better than a "city car" and may be interested in these worse roads too.
For bikes I am not sure what kind of bike OsmAnd tries to present, so for safety I allow all smoothness except "impassable".  	Because even "very_horrible" is defined to be passable by mountain bike.

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -98,8 +98,13 @@
 				<select value="-1" t="surface" v="pebblestone"/>
 				<select value="-1" t="surface" v="sand"/>
 				<select value="-1" t="surface" v="wood"/>
+				<select value="-1" t="smoothness" v="very_bad"/>
 			</if>
 
+			<select value="-1" t="smoothness" v="impassable"/>
+			<select value="-1" t="smoothness" v="very_horrible"/>
+			<select value="-1" t="smoothness" v="horrible"/>
+			
 			<select value="-1" t="motorcar" v="no"/>
 			<select value="-1" t="motorcar" v="agricultural"/>
 			<select value="-1" t="motorcar" v="forestry"/>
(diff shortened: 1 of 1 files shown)


vshcherb:
In some countries that's only the option to reach destination, I will invite from russian community to get feedback about it. There is no doubt that we should deprioritize but disallow is not a good option I think

aceman444:
Sure, please get some more opinions. We can also make it dependent on the avoid_unpaved option.

aceman444:
Hi, so were you able to get any feedback? Or do we allow all smoothness if avoid_unpaved is NOT selected?

xmd5a2:
Cyclists can get through almost all the places where a pedestrian can pass. So in theory cyclist can get off the bike and turn into pedestrian. Wiki says that smoothness=impassable  means that no wheeled vehicle can pass through way (but pedestrian can). Completely disabling smoothness=impassable for cyclist is a bad idea.

xmd5a2:
Please have a look at these ways http://overpass-turbo.eu/s/eod
If we make them non-routable these settlements can't be accessed at all.

aceman444:
OK, I understand. What about the car profile?

xmd5a2:
Disabling car routing through smoothness=impassable leads to the fact that the village Batolajy will be unattainable http://www.openstreetmap.org/way/169490469#map=15/-15.6748/47.0709
I think better not to disable routing at all but lower priority as far as possible. Same for horrible and very_horrible.

aceman444:
But then, is that correct tagging? Why is that road marked impassable, if a car can pass over it?

What about allowing such roads only when "avoid_unpaved" is NOT set?

xmd5a2:
Road to Batolajy contains seasonal=yes also. So this road is not completely impassable all the time. Until we do not support such conditions it is better not to disallow routing on such roads but lower priority.

xmd5a2:
Merged manually ba15b64a92d24c1d326cb046c5046df16e8c3ac8

aceman444:
Great, thanks!

PR #348, opened 2015-09-12
Added Paediatrics 

morray:
Added Paediatrics to make it easier to find these when travelling


xmd5a2:
Phrases in OsmAnd-resources/poi/phrases are copy from https://github.com/osmandapp/Osmand/tree/master/OsmAnd/res/values-**/phrases.xml
Please add translations to these files directly of use weblate.

PR #349, opened 2015-09-13
Update Czech TTS

vmonkey:
[Please do not merge yet]

+ Corrected declension of kilometry/kilometrů (kilometers)
+ Corrected declension of hodiny/hodin (hours)
+ Corrected declension of minuty/minut (minutes)
+ Corrected declension of dva/dvě (two)
+ Corrected missing spaces
+ Added support for Attention: type [speed_camera, border_control, railroad_crossing, traffic_calming, toll_booth, stop, pedestrian_crosswalk]
+ Generated updated voice files using google TTS
+ Corrected word order in time.ogg
+ Translated missing strings

If possible I would appreciate someone checking if it does really work as intended as it is not possible to check all possible situations. [However tested via developer plugin - both prerecorded and TTS]


aceman444:
Hi, in the patch you seem to also have changed the config.p file for the non-TTS (prerecorded) voices? Have you also changed something in the prerecorded .ogg files so that you had to attach them all here?

vmonkey:
Hi,
all the rules for non-TTS (prerecorded) voices are stated in the ttsconfig.p file too. Because I wanted to implement all the above changes into prerecorded voices I ran the "gen_voice.sh" which generates config.p and also all of the *.ogg files. As I didn't want to check all the ogg files one by one to see what had changed, I uploaded them all. New files are eg. 2_minutes.ogg and 3_4_minutes.ogg etc...

aceman444:
Which Google TTS does support Slovak?

vmonkey:
Google translate, see for example http://translate.google.com/translate_tts?ie=UTF-8&total=1&idx=0&client=t&tl=sk&q=Ofici%C3%A1lna%20str%C3%A1nka%20%C3%9Aradu%20vl%C3%A1dy%20Slovenskej%20republiky.

aceman444:
Thanks, I now understand what you do and that it is possible to generate the ogg files and config.p from the ttsconfig.p. Thank you fro your hints. I now see why it is needed to upload all the files, the voice used to create them may be different from the one in the old files. The same happens in the Slovak Google TTS.

aceman444:
Maybe you can also add the feature:
% (X) Announcement when user returns back to route

string('back_on_route.ogg', 'you have returned to route').

See e.g. the English tts.

aceman444:
You may also add 
string('and_arrive_poi.ogg', 'and pass POI ').
string('reached_poi.ogg', 'you are passing POI ').

vmonkey:
Thanks, I had to forget about these, I will add them soon. Regarding the back_on_route it seems to be implemented only as string('back_on_route.ogg', 'some_text'') but there is no "rule" for it. Is it OK?

vmonkey:
@sonora You seem to add this commit https://github.com/osmandapp/OsmAnd-resources/commit/9ab2fb15e4db125347a38bd5d7c41527d4d86478 regarding (then) (bear_left/right). Would it be possible to explain how to use it (possibly with an example)? Thank you very much!

sonora:
I am actually not sure... unless I totally misundersand, in English it
makes no difference. The instruction after "then"I s simply meant to hint
to the driver that the successive turn would be right or left, so he can
prepare for this (pick a good lane if there are several.) But I know of no
example in any language with a good difference between "bear"  and "keep"...

vmonkey:
Thanks, that was just my misunderstanding. What I was looking for was that in the Czech language we use a different word order if keep_left (or right) is 'standalone' or after 'then', ie:

Držte se vlevo. [keep_left] OR ...potom se držte vlevo. [... then keep_left] -- "se" is in different place.

Is there a way to implement this? If there is not, I will just upload the changes I have already made.

aceman444:
Yes, this is also a problem in Slovak. "potom, drzte se vlevo" is incorrect. And "prepare to keep left" is another version again. But 'left_keep_p' is already in the file, but I am not sure if it is used. I think we need some more samples in the Testing of voice commands in the dev plugin. Can we add some more in a new issue?

sonora:
Sure, I can add any test which make sense! My problem is that I do not
understand the problem ...    :-)   likely because I does not exist in
English  ... :-)

So if the issue is that
- after a "then" you would need another text (keep_left/right.ogg) than
- without the preceding "then" (bear_left/right.ogg)
then you could probably just define this with conditions when you build the
phrases in the ttsconfig?

But maybe I still misunderstand, so  you would have to go the extra mile t
try to explain it ... :-)

Thx - Hardy

aceman444:
Exactly, after "then" we need a new string, e.g. left_keep_t, because left_keep_p is already used after "prepare". Maybe it is exactly the string bear_left, but you also couldn't say how to use it.
There is only a rule like 'bear_left(_Street) -- ['left_keep.ogg'].' in the ttsconfig.p which we do not understand what it does. So if you could add a phrase using this command in the dev plugin, we could try it out.

sonora:
Let me explain again, I think we have all you need, you may just not be
aware of it:

There are 2 different types of commands passed to the tts engine, in
English I map them both to sound as "keep", so it is not obvious for you
these are 2 different commands:

bear_left(Street) and bear_right(Street) are triggered as brief
"turn-after-next" preparation. These are always _only_ used in connection
with "then". In the voice prompt test activity, the test prompts for these
are  (switch to English to see it) "In 370m turn slightly right onto 'F23'
'Main Street', then bear left", and "In 450m turn sharply left onto
'Hauptstrasse', then bear right"). Again, please note that in English tts,
the 2 "bear" commands are mapped to sound as "keep", but you can change
this in other languages..

And secondly there is an actual turn type called "left_keep" and
"right_keep". This is used to upcoming aid lane selection. The test voice
prompts for these are "After 1810m keep left ' '" and "In 400m keep left '
' then in 80m keep right 'A1'". For lack of better idea I also map to
"keep" in English tts, just like "bear".

Does this help?

Best, Hardy

aceman444:
I have found a single language that is using bear_* and that is "et". I will try to implement what they do at  https://github.com/osmandapp/OsmAnd-resources/blob/master/voice/et/ttsconfig.p#L223 and see how it works. Albeit I can only try the prerecorded voices, not TTS. But I will try the 2 test voice prompts, which I do see in dev plugin.

What does the Street mean in the bear_left(Street) command? Is there supposed to be street name put when using real TTS?

sonora:
Yes, a street name is being passed, so you can use it by calling (Street)
or ignore it by calling (_Street).

sonora:
PS: I have adjusted the comments in the English ttsconfg to better explain
the "keep" vs. "bear" commands and how to use in languages where
differentiating them is desired. Hope this helps!

vmonkey:
@sonora Nice explanation, I finally understand how it works and also implemented the above (and it works as expected, thanks)! The word order also got better with an exception that the "bear_left/right" string should be used also after ['after.ogg', D], ie. in 'after 50 meters bear left/right'. Is that also possible?
And where (ie. in which file) are implementations of functions such as num_atom() and atom_concat()? I added proper declensions for distance (ie. in 'after some distance' the number should have been locative) but it currently works only for non(tts) as I need to change more things.

@aceman444 If the fact that you can't test real TTS is because of the lack of free Slovak TTS engine, you may try this (I am currently 'running' Czech through it): http://forum.xda-developers.com/showthread.php?t=2566795.

vmonkey:
@aceman444 As these are similar problem both in Czech and Slovak, there is no need to follow the Ethiopian language. The change that works is just to create new strings, eg. string('left_bear.ogg', 'se držte vlevo '), and then modify the bear_left command by replacing 'left_keep.ogg' by 'left_bear.ogg'. But as I have already written it does not however work for the case 'after some distance keep left'.

vshcherb:
First impression: my biggest concern with this pull request is recorded voice files. What voice engine was used?

vmonkey:
I plan to close this pull and when all things are finished, I will create two independent - one for TTS and one for prerecorded voices to make it more simple. The prerecorded files are (and will be again when all options are implemented) generated using Google translate TTS using gen_voice.sh script from repository. Is it ok?

vshcherb:
Yes it is ok, but if the quality is good enough.

sonora:
@vmoney, You can change all word order etc. in the ttscongif.p file itself,
it is completely handled by PROLOG, including all functions definitions
like atom, etc.

aceman444:
vmonkey, thanks a lot. I have used Vocalizer trial version in the past. The new version you linked seems to work great on the OsmAnd tts file. Thanks a lot.

PR #352, opened 2015-09-24
Fix time(Sec) when hours>0 and minutes=0

vmonkey:
Fix case when hours > 0 and minutes = 0 (eg. 2 hours, 0 minutes)
Behaviour before change: '2 hours minutes' for prerecorded and '2 hours 00 minutes' for TTS
Behaviour after change: '2 hours' for both options.'


vshcherb:
Thanks, the only concern if we do need update other languages.

vmonkey:
Yes, probably most of them are needed to update. I will check and post another pull with all of them.

sonora:
Thanks!  -Hardy

PR #353, opened 2015-09-24
Fix time(Sec) for all languages

vmonkey:
All languages should be fixed. I was not however sure with 'ja', so I added the fix only for the TTS.


aceman444:
What is this doing? :)

aceman444:
Is it when there is an exact number of hours and 0 minutes?

aceman444:
grr, it is written in the commit message to each file :)

vshcherb:
Thanks

PR #354, opened 2015-09-27
Update Czech TTS

vmonkey:
- Corrected declension of kilometry/kilometrů (kilometers)
- Corrected declension of hodiny/hodin (hours)
- Corrected declension of minuty/minut (minutes)
- Corrected declension of dva/dvě (two)
- Added missing spaces
- Added support for Attention: type [speed_camera, border_control, railroad_crossing, traffic_calming, toll_booth, stop, pedestrian_crosswalk]
- Corrected word order in time.ogg and keep_left/bear_left
- Translated missing strings
- Announcement of favorites, waypoints and pois along the route
- Announcement when user returns back to route
- Added proper declension of distance

Everything should be working fine -- tested by myself on drive and also by developer plugin.
Big thanks to @sonoma for telling difference between bear and keep.

### DIFF
--- a/voice/cs/ttsconfig.p
+++ b/voice/cs/ttsconfig.p
@@ -19,11 +19,14 @@ language('cs').
 % (X) special grammar: accusative/locative for distance measure
 % (X) special grammar: pturn
 % (X) SVOX workaound for wrong distance declination
-% (!) distance measure: meters / feet / yard support: feet/yard still needs proper acc/loc strings
+% (X) distance measure: meters / feet / yard support: feet/yard still needs proper acc/loc strings
 % (X) Street name announcement (suppress in prepare_roundabout)
 % (X) Name announcement for destination / intermediate / GPX waypoint arrival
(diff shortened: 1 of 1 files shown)


vmonkey:
@vshcherb I also tested generating voices, they are however better from ispeech than from Google TTS (tested generating them using online demo), I do not however have a developer account there. If it is possible for you to generate Czech voices from ispeech, I would be really thankful. Otherwise, I may generate voices at least using Google translate TTS with a little bit lower quality.

vshcherb:
The issue with ispeech, I used to buy words there, but after "they lost" my 5000 generated words,  I don't know if I really want to buy it again. I remember Google TTS was not great, may be it is improved, but if you know a reasonable service please let me know. The infrastructure (scripts) should be ready to download bulk, diffs and so on, I just don't know what is the best engine.

vshcherb:
Here is an API http://www.ispeech.org/api, please let me know if you are interested, I can buy new credits and send you a private key.

PR #357, opened 2015-10-01
Update gen_voice.sh

EgorKin:
change version(102) to version(103) because all of ttsconfig.p files consist only version(103)
change {t::-4} to {t%????} fixed string manipulation for mingw32


aceman444:
Not really, there are still a ton of ttsconfig.p files that have not upped the version to 103, see https://github.com/osmandapp/OsmAnd-resources/search?utf8=%E2%9C%93&q=%22version%28102%29%22&type=Code . Maybe you could make the script support both versions?

DmitryEgorkin:
Done. Thanks for your comment.

EgorKin:
Something bad with my pull request? May be I should change something? Current gen_voice.sh won`t work with many langs with 103 version.

PR #360, opened 2015-11-06
Voice-only guidance files for Sardinian (sc)

l2212:
I couldn't find any tts engine that supports this language, right now.


vshcherb:
What language engine did you use ?

l2212:
Balabolka with an italian voice, Microsoft Elsa Desktop (the one most "compatible" with Sardinian) and adapting the input words to try to obtain the correct sounds.

vshcherb:
My only question would be how we are going to support new sounds in case, will it be ok to contact you if there are some tiny changes? If that's ok could you please put in comment somewhere that information what's been used.

l2212:
For me it's fine, I think. For information you mean my e-mail account?

l2212scr@yahoo.it

vshcherb:
I hope I could reach via Github and you check at least once per 3-4 months if main tts english file has changed.
Thanks for contribution!

l2212:
Ok, I will try to remember it. Thanks to you for your great work!

l2212:
Hello @vshcherb ,
one Sardinian University created a prototype, with some Sardinian voices for MARY TTS:

http://sintesa.eu/en/prototipo/

Could that be fine? Does OsmAnd supports that engine?

sonora:
Based of these comments https://github.com/marytts/marytts/issues/318 it should be possible - definitely worth a try.

l2212:
@sonora Cool! What should I do?

l2212:
Also, I've seen that the voice files are still not in the "downloads" list. Why?

PR #361, opened 2015-11-14
Update config.p

amirahmadMohamadi:
a typo is fixed in tts texts
also it seems recorded voices are not available in the software for Farsi.


sonora:
Just to make sure:, before we commit:  Are you sure it is a typo which needs to be corrected, right? In some cases strings in TTS file re misspelled on purposes to improve pronunciation ...

amirahmadMohamadi:
yes in fact I found it when I listened to the tts voice

sonora:
Merged now, please test. Thanks!

PR #362, opened 2015-11-15
Update ttsconfig.p

amirahmadMohamadi:
typo fixed


amirahmadMohamadi:
yes in fact I found it when i listened to the tts output voice

PR #365, opened 2015-12-01
Changed translation for poi_chemist, which is not "Farmacia" in Italian.

niccolo-r:
This is because adding a "Farmacia" POI in OsmAnd should insert amenity=pharmacy, not shop=chemist.
In Italy there are not "Chemist" shops, the more similar kind of shop is the "mesticheria", which sells various goods for housekeeping and personal care.


xmd5a2:
I merged it, but look at
https://github.com/osmandapp/OsmAnd-resources/blob/master/poi/phrases/README.txt

PR #366, opened 2015-12-23
bolivia regions (srtm & hillshade)

51114u9:
Split srtm & hillshade files per regions of Bolivia

### DIFF
--- a/countries-info/regions.xml
+++ b/countries-info/regions.xml
@@ -1298,7 +1298,7 @@
 				<region map="no" name="pando" translate="Pando;entity=relation"/>
 				<region map="no" name="potosi" translate="Potosí;entity=relation"/>
 				<region map="no" name="santa-cruz" translate="Santa Cruz;entity=relation"/>
-				<region map="no" name="tarija" translate="Tarija;ntity=relation"/>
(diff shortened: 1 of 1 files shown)


vshcherb:
Bolivia is rather small now and doesn't need to be split in 10 regions. Also we will need to create corresponding boundaries.

51114u9:
Well, many people can not download these large files with very low Internet connection. Do you know any mechanism to divide into smaller parts?

robertogeb:
You can create your own maps with OsmAndMapCreator. You can download the
raw map from http://download.geofabrik.de/ and then cut the piece you want
by using osmconvert or download the piece you want selecting it from
http://extract.bbbike.org/.

2015-12-30 3:33 GMT+01:00 Marco Antonio <notifications@github.com>:

> Well, many people can not download these large files with very low
> Internet connection. Do you know any mechanism to divide into smaller parts?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/osmandapp/OsmAnd-resources/pull/366#issuecomment-167920873>
> .
>



-- 
Saludos,

     Roberto

PR #368, opened 2016-01-01
Update phrases.xml (spanish)

OpenSourceForEver:


xmd5a2:
Please read https://github.com/osmandapp/OsmAnd-resources/blob/master/poi/phrases/README.txt and use https://hosted.weblate.org/projects/osmand/phrases/

OpenSourceForEver:
I did it because the link showed on  https://github.com/osmandapp/OsmAnd-resources/blob/master/poi/phrases/README.txt is not correct

Sorry

PR #369, opened 2016-01-07
Renaming syntax with a more cordial

franco999:
https://es.wikipedia.org/wiki/Wikipedia:%C2%BFT%C3%BA_o_usted%3F

### DIFF
--- a/voice/es/ttsconfig.p
+++ b/voice/es/ttsconfig.p
@@ -35,18 +35,18 @@ string('route_calculate.ogg', 'Ruta recalculada ').
 string('distance.ogg', ', distancia').
 
 % LEFT/RIGHT
-string('prepare.ogg', 'Prepárate para').
+string('prepare.ogg', 'Prepárese para').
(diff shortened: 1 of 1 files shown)


robertogb:
I don't agree with this request. It is about making the human-machine interface more distant by establishing a barrier owner-servant. Right now, the language used is more friendly. It was discussed two years ago and it was agreed not to change. The user franco999 is changing the user interface in Spanish claiming that it is Spanish neutral versus the former Spanish Spain version, but for me this is Spanish South America. There is a Spanish Argentina version and they can change that or create a new one for other Spanish version.

sonora:
@franco999: Would this be a solution, to place this pull request again for es-AR?

franco999:
Sorry, I was wrong to place comment. The commentary is in the description of the commit.

sonora:
Since there was no further feedback here, I have imported this as es-ar for now, where it can be tested, and we can decide if to keep this and under what language code.

PR #371, opened 2016-02-29
Deleted motorway_link from car navigation

hakuchi:
I think whith this little modification [#1202](https://github.com/osmandapp/Osmand/issues/1202) is fixed.
The motorway_link ending/coming from a motorway, so I think, it is not necessary to add this to the routing.xml

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -72,9 +72,6 @@
 			<select value="-1" t="highway" v="motorway">
 				<if param="avoid_motorway"/>
 			</select>
-			<select value="-1" t="highway" v="motorway_link">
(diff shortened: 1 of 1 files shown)


vshcherb:
Let's give it a try. Looks promising.

aceman444:
I don't think this fix is correct. You (or user) can disable routing over various types of roads in different scenarions/navigation modes (e.g. trunks or tracks). But the roads are physically there and the user will see them in real life, so the engine must take them into account when navigating/displaying the exits or voice commands.

This can't be fixed by removing only motorway_link from routing.xml. It is an incorrect solution. The internal engine must be fixed.

PR #372, opened 2016-03-11
fr : rephrasing of "Location lost/recovered" to "Localisation perdue/retrouvée"

breversa:
"Location lost/recovered" was previously translated to "GPS signal lost/recovered".

However :
- GPS is only one of the several existing satellite positionning systems (GLONASS, Beidou, later Galileo) 
- Satellite positionning is not the only existing positionning : cell towers, wifi hotspots, later Bluetooth and other technologies (RFID/NFC tags ?) can be used as well
- Location can be lost even though GPS signal (or signals) has not been lost

For those reasons, I propose a more correct translation of "Localisation perdue/retrouvée"

Also :
- disambiguation of "serrez/restez"
- "route_calculate.ogg" is played when calculation is over, hence "Itinéraire recalculé"

### DIFF
--- a/voice/fr/ispeech_fr.csv
+++ b/voice/fr/ispeech_fr.csv
@@ -1,5 +1,5 @@
-route_is.ogg,l'itinéraire fait  
-route_calculate.ogg,recalcul de l'itinéraire
+route_is.ogg,L'itinéraire fait  
+route_calculate.ogg,Itinéraire recalculé
 distance.ogg,  l'itinéraire fait 
 prepare.ogg,préparez vous à 
 after.ogg,dans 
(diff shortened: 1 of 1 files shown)


vshcherb:
This pull-request will not changed recorded files and these files should not be modified manually

breversa:
That's what I feared.

What should be done then ?

2016-03-11 17:37 GMT+01:00 vshcherb <notifications@github.com>:

vshcherb:
Not much we don't have possibility to regeneeate ispeech voice, so we need to fully regneerate voice package then.

breversa:
Thanks for your reply, but I don't get it. :-/

What IS possible, and what ISN'T ?

2016-03-12 19:39 GMT+01:00 vshcherb <notifications@github.com>:

vshcherb:
Without regeneration of all voice recordings, it is not possible today. We don't have regeneration setup.

breversa:
So basically, we used to have that capability, but no longer ? How come ?

2016-03-14 13:42 GMT+01:00 vshcherb <notifications@github.com>:

PR #374, opened 2016-04-05
Update regions.xml

franco999:
Edit the code to divide the standard map of Argentina in the various sub-regions smaller like France or Germany. I don't know much about this language, please correct what is wrong.

### DIFF
--- a/countries-info/regions.xml
+++ b/countries-info/regions.xml
@@ -1290,7 +1290,7 @@
 
 		<region type="continent" name="southamerica" translate="=South America" inner_download_suffix="southamerica" boundary="no" poly_extract="south-america">
 			<!-- <region name="argentina" hillshade="no" srtm="no" inner_download_prefix="$name" lang="es"> -->
-			<region name="argentina" inner_download_prefix="$name" hillshade="no" srtm="no" lang="es" poly_extract="south-america">
+			<region name="argentina" inner_download_prefix="$name" hillshade="no" srtm="no" lang="es">
 				<region type="hillshade" name="ciudad-de-buenos-aires" translate="name:en=Autonomous City of Buenos Aires;entity=relation"/>
 				<region name="buenos-aires" translate="Buenos Aires;entity=relation"/>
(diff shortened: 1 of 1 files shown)


vshcherb:
@xmd5a2 Do we have poly files for these regions?

countries-info/regions.xml
```diff
@@ -1289,31 +1289,32 @@
 		</region>
 
 		<region type="continent" name="southamerica" translate="=South America" inner_download_suffix="southamerica" boundary="no" poly_extract="south-america">
-			<region name="argentina" hillshade="no" srtm="no" inner_download_prefix="$name" lang="es">
-				<region map="no" name="buenos-aires" translate="Buenos Aires;entity=relation"/>
+			<!-- <region name="argentina" hillshade="no" srtm="no" inner_download_prefix="$name" lang="es"> -->
+			<region name="argentina" inner_download_prefix="$name" hillshade="no" srtm="no" lang="es" poly_extract="south-america">
```
xmd5a2: No need to specify poly_extract here.

xmd5a2:
@vshcherb Yes, we have polys for all regions in regions.xml.

franco999:
@xmd5a2 Done, poly_extract. any other changes needed?

vshcherb:
I'm fine with merging it though we also need to split China, Sweden, Australia, Indonesia, Mexico, Turkey, Romania, Switzerland, Brazil, Norway and even Denmark, Belgium but that analogy as well.

aceman444:
Why split those countries? Is there some file size criterion?
Argentina as such was not a big file to download (160MB?). But maybe it makes sense to split it as it is geographically large, and most people do not need it all as they do not cross it from north to south.
Switzerland, Denmark or Belgium are so small geographically, that you can cross them in hours. So unless their files are in hundreds of MBs, there is not much reason to. Splitting countries is harder for the users to manage (for users foreign to that country), as they need to find out which regions they need (OsmAnd should help them via the on-map info, but it doesn't work well).

On the other hand, if you look at recent Maps.me, they have also split very small countries into regions.

PR #375, opened 2016-04-13
Better bicycle routing

Posteinwurf:
Dear all,

this is a suggestion for an improved bicycle routing. It is based on many, many days of work, tests and gradual improvements. It has been field tested extensively in Berlin, Hamburg and Cologne (and tested virtually in France, Poland and Germany, with short and long distances).

After all those gradual improvements, I think it now works like magic: In almost any test I did, it found a better cycle route than the original version. A route with less car traffic (= less noise, less danger, less pollution). In most cases it will find great routes trough calm residential areas, parks, forests or along waterways. Still, it will find a fast route: It heavily prefers paved paths, tracks and roads.

Some observations:
a) I recommend to check the box "Less Traffic Signals - Less Main Roads". Then it will find even more hidden byways without cars.
b) With that box checked, the route will be 5 to 15 percent longer than in the current version, but in spite of the additional meters, it will normally still be faster (because it avoids traffic lights).
c) If that box is unchecked, the route will be a compromise between the old version and the new version.
d) At first glance, the coefficients might seem extreme, but many tests have shown that they are balanced: Even the new version will use main roads regularly, but only if they provide a real shortcut.
e) Please note that the new coefficients do not reflect real travel time at all. Instead, they only indicate preferences.

Please feel free to comment. I would also be glad if some people could test it! Apart from "real world tests" you can also do quick "virtual tests" with bike routes that you know well.

Greetings from Berlin

Malte

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -466,14 +466,14 @@
 			<select value="1"  t="bicycle" v="use_sidepath"/>
 			<select value="1"  t="bicycle" v="dismount"/>
 
+			<select value="-1" t="vehicle" v="no"/>
 			<select value="1" t="vehicle" v="agricultural"/>
 			<select value="1" t="vehicle" v="forestry"/>
 			<select value="1"  t="vehicle" v="yes"/>
 			<select value="1"  t="vehicle" v="permissive"/>
 			<select value="1"  t="vehicle" v="designated"/>
 			<select value="1"  t="vehicle" v="destination"/>
-			<select value="-1" t="vehicle" v="no"/>
-
+			
 			<select value="-1" t="access" v="no"/>
 			<select value="1" t="access" v="agricultural"/>
 			<select value="1" t="access" v="forestry"/>
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -564,7 +564,7 @@
 			<select value="17" t="highway" v="path"/>
 			<select value="23" t="highway" v="living_street"/>
 			<select value="8" t="highway" v="pedestrian"/>
-			<select value="15.5" t="highway" v="footway"/>
+			<select value="13" t="highway" v="footway"/>
 			<select value="10"  t="highway" v="byway"/>
 			<select value="2" t="highway" v="platform"/>
 			<select value="2" t="highway" v="services"/>
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -607,7 +607,7 @@
 			<select value="1.0" t="tracktype" v="grade3"/>
 			<select value="0.7" t="tracktype" v="grade4"/>
 			<select value="0.5" t="tracktype" v="grade5"/>
-			<select value="0.7" t="surface" v="cobblestone"/>
+			<select value="0.65" t="surface" v="cobblestone"/>
 			<select value="1.5" t="cycleway" v="lane"/>
 			<select value="1.5" t="cycleway" v="opposite_lane"/>
 			<select value="1.1" t="cycleway" v="share_busway"/>
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -559,15 +559,15 @@
 			<select value="23.5" t="highway" v="residential"/>
 			<select value="22.0" t="highway" v="cycleway"/>
 			<select value="23" t="highway" v="unclassified"/>
-			<select value="15.5" t="highway" v="service"/>
+			<select value="11" t="highway" v="service"/>
 			<select value="19.0" t="highway" v="track"/>
 			<select value="17" t="highway" v="path"/>
 			<select value="23" t="highway" v="living_street"/>
-			<select value="11" t="highway" v="pedestrian"/>
+			<select value="8" t="highway" v="pedestrian"/>
 			<select value="15.5" t="highway" v="footway"/>
 			<select value="10"  t="highway" v="byway"/>
-			<select value="11" t="highway" v="platform"/>
-			<select value="13" t="highway" v="services"/>
+			<select value="2" t="highway" v="platform"/>
+			<select value="2" t="highway" v="services"/>
 			<select value="2" t="highway" v="bridleway"/>
 			<select value="14.5"/>
 		</way>
@@ -794,7 +794,7 @@
 			<select value="0.05" t="foot" v="private"/>
 			<select value="0.05" t="foot" v="destination"/>
 
-			<!-- Q? access=destination should not be used foot or what is the proposed schemeß
+			<!-- Q? access=destination should not be used foot or what is the proposed scheme
 			http://www.openstreetmap.org/way/173188624#map=18/39.56805/2.64846
 			should be 
 			<select value="0.05" t="access" v="private"/>
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -639,13 +639,17 @@
 			<select value="1.1" t="surface" v="compacted"/>
 			<select value="1.1" t="surface" v="fine_gravel"/>
 			<select value="1.0" t="surface" v="wood"/>
+		<if param="avoid_unpaved">
+			<select value="0.65" t="highway" v="footway"/>
+			<select value="0.5" t="highway" v="track"/>
+			<select value="0.45" t="highway" v="path"/>
+		</if>
 			<select value="0.76" t="highway" v="footway"/>
 			<select value="1.0" t="highway" v="unclassified"/>
 			<select value="1.0" t="highway" v="road"/>
 			<select value="1.0" t="highway" v="service"/>
 			<select value="0.75" t="highway" v="track"/>
 			<select value="0.6" t="highway" v="path"/>
-
 			<select value="0.5" t="highway" v="motorway"/>
 			<select value="0.5" t="highway" v="motorway_link"/>
 			<select value="0.5" t="highway" v="trunk"/>
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -466,13 +466,13 @@
 			<select value="1"  t="bicycle" v="use_sidepath"/>
 			<select value="1"  t="bicycle" v="dismount"/>
 
-			<select value="1" t="vehicle" v="no"/>
 			<select value="1" t="vehicle" v="agricultural"/>
 			<select value="1" t="vehicle" v="forestry"/>
 			<select value="1"  t="vehicle" v="yes"/>
 			<select value="1"  t="vehicle" v="permissive"/>
 			<select value="1"  t="vehicle" v="designated"/>
 			<select value="1"  t="vehicle" v="destination"/>
+			<select value="-1" t="vehicle" v="no"/>
 
 			<select value="-1" t="access" v="no"/>
 			<select value="1" t="access" v="agricultural"/>
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -425,15 +425,19 @@
 		</point>
 	</routingProfile>
 
-	<routingProfile name="bicycle" baseProfile="bicycle" restrictionsAware="true" minDefaultSpeed="7" maxDefaultSpeed="27"
-			leftTurn="0" rightTurn="0" followSpeedLimitations="false" onewayAware="true" heuristicCoefficient="1.5">
+	<routingProfile name="bicycle" baseProfile="bicycle" restrictionsAware="true" minDefaultSpeed="2" maxDefaultSpeed="28.5"
+			leftTurn="0" rightTurn="0" followSpeedLimitations="false" onewayAware="true" heuristicCoefficient="1.4">
 		<!-- <attribute name="relaxNodesIfStartDistSmallCoeff" value="2.5"/>  -->
 		<!-- New ROUTING API -->
-		<parameter id="avoid_ferries" name="Avoid ferries" description="Avoid ferries" type="boolean"/>
-		<parameter id="avoid_motorway" name="Avoid motorways" description="Avoid motorways" type="boolean"/>
+
+		<parameter id="avoid_traffic_signals" name="Less Traffic Signals - Less Main Roads" description="Avoid Traffic Signals" type="boolean"/>
 		<parameter id="avoid_unpaved" name="Avoid unpaved roads" description="Avoid unpaved roads" type="boolean"/>
+		<parameter id="avoid_ferries" name="Avoid ferries" description="Avoid ferries" type="boolean"/>
 		<parameter id="avoid_stairs" name="Avoid stairs" description="Avoid stairs" type="boolean"/>
 		<parameter id="avoid_borders" name="Avoid border crossing" description="Avoid crossing a border into another country" type="boolean"/>
+		<parameter id="avoid_motorway" name="Avoid motorways" description="Avoid motorways" type="boolean"/>
+
+
 
 		<way attribute="access">
 			<select value="-1" t="osmand_change" v="delete"/>
@@ -450,6 +454,9 @@
 				<if param="avoid_stairs"/>
 			</select>
 
+			<select value="-1" t="highway" v="elevator"/>
+			<select value="-1" t="highway" v="construction"/>
+			<select value="-1" t="motorroad" v="yes"/>
 			<select value="-1" t="bicycle" v="no"/>
 			<select value="1"  t="bicycle" v="yes"/>
 			<select value="1"  t="bicycle" v="permissive"/>
@@ -459,17 +466,17 @@
 			<select value="1"  t="bicycle" v="use_sidepath"/>
 			<select value="1"  t="bicycle" v="dismount"/>
 
-			<select value="-1" t="vehicle" v="no"/>
-			<select value="-1" t="vehicle" v="agricultural"/>
-			<select value="-1" t="vehicle" v="forestry"/>
+			<select value="1" t="vehicle" v="no"/>
+			<select value="1" t="vehicle" v="agricultural"/>
+			<select value="1" t="vehicle" v="forestry"/>
 			<select value="1"  t="vehicle" v="yes"/>
 			<select value="1"  t="vehicle" v="permissive"/>
 			<select value="1"  t="vehicle" v="designated"/>
 			<select value="1"  t="vehicle" v="destination"/>
 
 			<select value="-1" t="access" v="no"/>
-			<select value="-1" t="access" v="agricultural"/>
-			<select value="-1" t="access" v="forestry"/>
+			<select value="1" t="access" v="agricultural"/>
+			<select value="1" t="access" v="forestry"/>
 			<select value="1"  t="access" v="yes"/>
 			<select value="1"  t="access" v="permissive"/>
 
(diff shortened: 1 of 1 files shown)


Posteinwurf:
This is a typical example of how the new version changes the preferences. In this case, the old version routes along main roads. They have got cycle tracks, but still they are not ideal (many traffic light stops, dangerous, loud, ugly). This is the old version's route for the last three kilometers:

even older


The new version finds a better route:

old


In this case it is a bike and hiking way that runs along a waterway, as you can see in this closeup and photo:




In OSM data, this way is marked as a "path" (which - at first glance - seems to indicate that it is not to be preferred). But this path has strong indicators that it will probably be better than the main road (even a main road with a bike track): The path is marked with "bicycle=designated" and "surface=asphalt". This combination ist found very often with paths that run through parks, residential areas or along waterways.

Greetings from Berlin

Malte

Posteinwurf:
For reference, I also provide this matrix: It shows how the new routing.xml will weigh the main combinations (that are found often in OSM maps)

matrix

vshcherb:
Hi Looks really cool and deep analysis made. I'm very convinced to update routing.xml the only question I would like to discuss how to automate testing these routes. Imagine situation 3 months later it will be another pull-request but we certainly don't want to ruin your use cases that you well tested. 

We have today an "artificial" map (https://github.com/osmandapp/OsmAnd-resources/blob/master/turn_lanes_test.osm) where we add some pieces of real map and run routing algorithm. I'm questioning would it be possible to add some tests or specify which parts could be added to show previous and currently selected route. Keeping tests as simple as possible of course is the best. 

This is an open question, just to kick off a discussion.


Best Regards,
Victor

Posteinwurf:
Dear Victor,

thanks for the fast and positive reply!

There are about 10 "standard tests" that I have run with each new set of coefficients. To run the tests, I simply used Osmand (in Berlin, Hamburg, Paris and in the countryside). I used areas that I know quite well (so I knew which route would be optimal). Maybe such a "real world test" is best, because it can also take into account OSM data inaccuracies. I understand too little of the "artificial map" to tell if my "handmade" tests can be integrated into that map...  But I would be willing just to run my "handmade" tests again, if there are further improvements.

All the best,

Malte

P.S.: I have some small _additional _ideas in mind, but to implement these, I need to ask some questions. Shall I already put these questions here, or do you first want some time to work through the many lines of input that I have already given?

aceman444:
Hi, thanks for the work! But why changing the access permissions? Why do you allow bikes into vehicle=no/agricultural/forestry? Or do you penalize those roads later on, e.g. with low speed due to pushing the bike?
  
-			<select value="-1" t="vehicle" v="no"/>
-			<select value="-1" t="vehicle" v="agricultural"/>
-			<select value="-1" t="vehicle" v="forestry"/>
+			<select value="1" t="vehicle" v="no"/>
+			<select value="1" t="vehicle" v="agricultural"/>
+			<select value="1" t="vehicle" v="forestry"/>

Posteinwurf:
I really enjoyed doing this work! Thank you for the feedback. I have used three sources: 

 

1.) Real world tests.

 

2.) Mapping guidelines in the OSM Wiki.

 

3.) Research in OSM maps: Which tags are actually used for highways that are good for cycling? And also: Which tags are often used wrongly?

 

This (frequent wrong usage) is exactly the case with vehicle=no : It is often used for highways where only the motorised vehicles are forbidden. See for example: https://www.openstreetmap.org/way/396621211 (it is even explicitly marked “bicycle=yes”). Those areas which are really forbidden for bicycles (like military terrain), would always be marked “access=no”. 

 

Agricultural / forestry: If they are paved, these are normally very attractive highways for bicycles. See for example this one from my place of birth: https://www.openstreetmap.org/way/6269655 (It is also explicitly marked “bicycle=yes”).

 

Greetings from Cologne,

 

Malte

aceman444:
But I am not sure we can just assume people tag incorrectly. Actually in the cases you have shown, the tagging is exactly correct. There is a broader vehicle=no that is then overriden for bikes with bicycle=yes. OsmAnd's routing.xml picks up the bicycle=yes first so the road is allowed to drive by bike. You do not need to remove  [select value="-1" t="vehicle" v="no"/] to get that result.

So if there is no bicycle=yes on a vehicle=forestry/argicultural road, bikes are NOT allowed. That is defined at http://wiki.openstreetmap.org/wiki/Key:access. But there may be per-country exceptions.

Posteinwurf:
Oh yes, you are right of course! I forgot that the bicycle=yes comes first and then overrules the “access=no”. I have changed it back to the original.

Still, I would rather leave the permission for vehicle=agricultural and vehicle=forestry. These are normally minor roads, so they tend to be mapped less accurately than main roads: Often it is not explicitly marked that they are allowed for bicycles. But normally, bicycles are tolerated on these roads (or at least people would not mind you using them).

This is one example where someone just forgot to add bicycle=yes: There is a NOTE IN GERMAN that the track is allowed for bicycles, but no corresponding tag. www.openstreetmap.org/way/292232119

aceman444:
I don't see you fixed the patch. You only moved the access=no but not removed the changes to vehicle=agricultural/forestry.

The minor roads you mean are marked highway=track (or similar). If there is no explicit access tags, OsmAnd allows cycling on them.

If there is an explicit access tag, we must obey it. I don't think we have to assume the mapper did it wrong. It is better if OsmAnd is more conservative and just obeys the tag. The tag is defined on the wiki so mappers should use it correctly.

You can fix the sample road and add the bicycle=yes. If it is allowed in all of Germany that bikes can go on *=agricultural marked roads, then that is a case for per-country defaults (which are not yet implemented in OsmAnd :()

Posteinwurf:
Dear aceman,

I see your point. I think the conservative approach has advantages, but also disadvantages. 

One advantage is that you minimise the risk to put people in conflict with traffic rules. But I think this risk is really small in rural areas. I have done thousands of kilometers with my bike (in Germany, Poland, Austria, France, Lithuania, Italy, Cameroon), but I never met anyone who objected to me using ways that were marked "agricultural" or "forestry". I think, even if you would ask the authorities they would often reply "we just did not think about cyclists", rather than "oh it is strictly forbidden to cycle there".

ekopie

My main idea is to include as many good options into the routing as possible. I have spent much research on the question: How can we get the optimal result at this very moment - with the deficiencies that still exist in the OSM data? Eliminating the deficiencies in the maps is no option, because there are too many, and in most cases you would have to do field research. 

Greetings from Berlin

Malte

aceman444:
I understand what you mean. But "nobody would really object" is not the same "it is 100% legal in that country". I don't think OsmAnd by default should allow potentially illegal routes.
You can of course use your own modified routing.xml file if you wish.

On the other hand, maybe you could add those forestry/agricultural roads as a toggleable option in Osmand, that would surely be fine for everybody. Or you could give them a high penalty (like value="0.1") so that OsmAnd takes such roads only as last resort and for short segments.

Posteinwurf:
Thank you for your ideas! I think that would be a great solution: Make a toggleable option. 

Suggestion: In order not to confuse the user with too many options, we could include the agricultural / forestry roads into the option "less traffic lights - less main roads" - and at the same time rename it in something like "more hidden byways (experimental)" or "even less main roads (beta)". 

If the team agrees, I will put it in the code.

xmd5a2:
Personally, I prefer the main roads because 1) in my city no cycling infrastructure 2) riding on the sidewalk at times slow and dangerous for pedestrians 3) often the main roads are the fastest way. I want to say that the bicycle routing must consider the preferences of the individual. I propose to make an option for my driving style - to give preference to the main road and not use the sidewalks.

Posteinwurf:
Sorry for replying to your comment so late (I have been in work projects).

If you tell me which city you live in, I could do some "virtual tests". Normally, the app should not take sidewalks at all. If there are no cycleways in your city, it will mainly take residential or minor roads, and sometimes main roads (for shortcuts). In the cities that I have tested so far, the preferences lead to a slightly longer route, but it is often faster, because it avoids traffic lights. But I would have to test your city to verify.

xmd5a2:
Hi. Thanks for response. I live in the city of Vladimir http://www.openstreetmap.org/?lat=56.131930330899&lon=40.4016545695468&mlat=56.131930330899&mlon=40.4016545695468&zoom=15

Posteinwurf:
Thank you! I have done some checks with overpass-turbo, to get an idea of the “layout” of your city. My first impression is, that the routing.xml would most often take one of the residential roads that are often marked with “asphalt”. Like this one www.openstreetmap.org/way/44682603 Are these generally good for cycling, in your city? 

 

There are indeed relatively few ways marked as “cycleways” or “bicycle=yes” or “bicycle=designated”. I found some in parks, like this one www.openstreetmap.org/way/414843826 

 

I will do some tests within Osmand. Maybe you could tell me some routes that you would normally take when cycling.  You could also do some tests yourself and see if the routes that the routing.xml finds suit you. In my own city, I have been positively surprised by the new routing.xml: Many times it found routes that were better than the ones that I had taken for months!

xmd5a2:
I've did some tests in my city. Your routing.xml prefer footways over primary, secondary and even tertiary. It is illegal in my country to drive through footways. Besides, it is much slower and more dangerous to pedestrians. That's why I and 90% of my friend bikers prefer main roads. At current state this routing.xml is not suitable for me and local realities. I think we should add more options to fill this demands.

Posteinwurf:
thank you for testing! 

 

I will downgrade the footways in the „normal“ mode (the mode with “less main roads” unchecked). This will help to make it suitable to everyone (I have tested in some cities, but mainly in Germany). Can you tell me which routes you have been testing (start/end point), so that I can find a reasonable value?

 

We could also add a third option “prefer main roads”, but I need help with that: Would it be possible to program a slider with three options? 1.) prefer main roads, 2.) Normal, 3.) Avoid main roads. Otherwise it would be a bit confusing: One option that toggles between “normal” and “prefer main roads”, and a second option that toggles between “normal” and “less main roads”.

xmd5a2:
That's good idea with slider (or may be dropdown menu). I think @vshcherb should pay attention to this issue.

xmd5a2:
I've tested these routes:
https://dl.dropboxusercontent.com/u/4624786/osm/Screenshots/Screenshot%20from%202016-05-12%2020%3A48%3A51.png
https://dl.dropboxusercontent.com/u/4624786/osm/Screenshots/Screenshot%20from%202016-05-12%2020%3A51%3A50.png
And this case is in reality much faster when using primary roads
https://dl.dropboxusercontent.com/u/4624786/osm/Screenshots/Screenshot%20from%202016-05-12%2020%3A52%3A47.png
That's because a relief (which we can't use for routing) but also because of big number of turns and of course footways.
May be we can reduce a number of left turns when using “prefer main roads” option? Often it's just deadly unsafe to turn left.

Posteinwurf:
Thank you for testing! I have saved your screenshots for future testing. I have downloaded the Oblast Vladimir map, so that I can include it in future tests.

 

Once the “three option slider” is implemented, I will include a “prefer main roads” mode into the routing.xml. I have no knowledge of the main program’s code, so I cannot do the slider myself. I have understood the logic of the routing.xml, but everything outside this little “sandbox” is unknown to me. J

vshcherb:
I don't see an issue with 3 type slider, you can define numeric parameter (like weight)
<gt value1=":param" value2="2" />
<le value1=":param" value2="1" />

Unfortunately there is no <eq value1 value2 > yet, though it should not be a problem to implement and possibly gt, le would be enough to do what is needed.

After many experiments with leftTurn, rightTurn , we supposed leftTurn=rightTurn should be equal and they are used in the routing, turn is detected on the turn angle.

xmd5a2:
I tried to add parameter like max weight but it is displayed in UI as regular on/off swith.

SlowBiker:
I prefer roads with less or no car traffic. Nearly all routes I got from OsmAnd I had to change by means of way-points to make it a route with less car traffic.
It would be nice, to have a switch to choose between fast routes and calm routes.

Posteinwurf:
maybe you would like to test the new bicycle mode that I proposed for Osmand?

All you have to do is to put an .xml file into the Osmand directory. Then, Osmand will use the new bicycle mode which avoids major roads (while still using paved roads most of the time)

I would be glad to have more people testing the new version. I could send you the .xml file via Email.

 

Malte

 

Von: SlowBiker [mailto:notifications@github.com] 
Gesendet: Sonntag, 14. August 2016 21:58
An: osmandapp/OsmAnd-resources
Cc: Malte; State change
Betreff: Re: [osmandapp/OsmAnd-resources] Better bicycle routing (#375)

 

I prefer roads with less or no car traffic. Nearly all routes I got from OsmAnd I had to change by means of way-points to make it a route with less car traffic.
It would be nice, to have a switch to choose between fast routes and calm routes.

SlowBiker:
Hello,

Yes, I would like to test the new bicycle mode.

You may send it to qx34898@yahoo.de

Best regards,

Klaus, from Munich, Germany



Am 25.08.2016 um 22:26 schrieb Malte:
> maybe you would like to test the new bicycle mode that I proposed for
> Osmand?
>
> All you have to do is to put an .xml file into the Osmand directory.
> Then, Osmand will use the new bicycle mode which avoids major roads
> (while still using paved roads most of the time)
>
> I would be glad to have more people testing the new version. I could
> send you the .xml file via Email.
>
>
>
> Malte

Posteinwurf:
Dear all, 

I have also tested the new routing.xml on a 1100 kilometer bike trip in rural France. 
I have added a modification to the "avoid unpaved" option. Now, also paths and tracks with unknown surface will be downgraded when this option is chosen.

All the best

Malte

vshcherb:
I'm closing this issue/pull request cause we have no power to support custom routing yet. We will be working on unit testing of bicycle routing. In parallel I would like to request to keep all your findings. Our first plan is to introduce "author view" routing, very similar concept to rendering style. So end user could choose offline routing.xml based on description of routing.xml.  https://github.com/osmandapp/Osmand/issues/3188

Posteinwurf:
Sorry, but I have put so much work into it. 

My suggestion: Just integrate my solution as the standard bicycle routing. For most users, it is much better.

 

 

 

Von: vshcherb [mailto:notifications@github.com] 
Gesendet: Donnerstag, 20. Oktober 2016 00:44
An: osmandapp/OsmAnd-resources
Cc: Malte; State change
Betreff: Re: [osmandapp/OsmAnd-resources] Better bicycle routing (#375)

 

I'm closing this issue/pull request cause we have no power to support custom routing yet. We will be working on unit testing of bicycle routing. In parallel I would like to request to keep all your findings. Our first plan is to introduce "author view" routing, very similar concept to rendering style. So end user could choose offline routing.xml based on description of routing.xml. osmandapp/Osmand#3188 <https://github.com/osmandapp/Osmand/issues/3188>

vshcherb:
To be sure your work is not lost and shared with other OsmAnd users! For now I put it as an extra file in repo https://github.com/osmandapp/OsmAnd-resources/blob/master/routing/routing-Posteinwurf.xml, it is not used yet, but it will be. At least it is visible downloadable and maintanable in some sense, please feel free to create pull-requests to it.

xmd5a2:
@Posteinwurf Hi. I slightly changed your bicycle routing and replaced one in routing.xml
https://github.com/osmandapp/OsmAnd-resources/commit/056de67b6ed6d6f0c08635be3d372501c631be9e
There is bike_type option where user can select "City bike", "Balance", "Fast bike". Routing preparation UI is about to be reworked but now it is accessible by "Route preparation screen -> Button with gear -> Navigation settings". "Fast bike" prefers main highways if no cycleways is available. "Balance" is your routing by refault. "City bike" is your "Less traffic signals" option.

Posteinwurf:
Dear xmd5a2,

Thank you very much. I think that is a great solution! With your idea, the slider is not necessary any more. People can just chose one of the three options. I am happy that you are integrating this into the code now. If you want, I can try to make the balanced mode more balanced. If I get it right, this will be part of the next updated then? When will it be uploaded?

At the moment, I would just suggest two changes: 

1.) I would rename "Fast bike" into "Prefer main roads" and "City bike" into "Avoid main roads". "City bike" is not self-explaining, and also, it rather depends on your cycling preferences than on the type of bike you are using.

2.) As a security measure, I would exclude motorways and motorroads entirely. As far as I can see, there is still an option "avoid motorways", and it is even unchecked (!) by default. But there are very few countries where it is allowed to cycle on motorways. 

Thank you very much and all the best!

Malte

P.S.: In the pedestrian mode, there is a  comment, that contains the German letter "ß". This letter should be deleted, because it leads to problems when viewing the xml file with browsers. The "ß" can be found at the end of this line: "<!-- Q? access=destination should not be used foot or what is the proposed schemeß"

xmd5a2:
@Posteinwurf
1) Partially agree, but "Prefer main roads" can be misunderstood as "Prefer ONLY main roads" while this routing prefer cycleways at most.
2)Agree. I'll fix it.

Posteinwurf:
Thank you, that sounds good. Great! 

Some fine tuning might be good for the balanced mode, so that people really get three entirely different modes, according to their personal preferences. I will take the time to look into it during the christmas holidays.

1.) You are right. Maybe we can find another phrase like „more main roads“ / “balanced” / “less main roads” Or: “include main roads” / “balanced” / “avoid main roads” Something like this.

Thank you for thinking about the traffic signal problem. Its true: If you go into the details, it gets complicated, also because some details are mapped in a different way in each region.

One problem that we could still solve: I think elevators are not recognised by Osmand, right? So sometimes, it would route using a public elevator (e.g. in the "Ostkreuz" train station in Berlin). It would be great to set a penalty for elevators. 

All the best,

Malte

Posteinwurf:
Dear Osmand Team,

 

thank you for releasing the new version with an improved bicycle routing. I could not really test it yet. But it is great that you are even starting to take elevation into account!!! I would like to suggest some tiny (but important) improvements:

 

1.)   I think the naming of the three preferences is misleading. At the moment it is called “safe”, “balanced” and “fast”. But it should rather be „more main roads“ / “balanced” / “less main roads”.  Because: a) If a cyclist avoids main roads, he will often be faster (because he avoids traffic lights as well). b) “Safe” does not tell the user what the preference actually changes. If you name it “less main roads”, it is clear what you get.

2.)   The choice “Autobahnen berücksichtigen” is quite misleading, at least in German (would be something like “take motorways into account” in English). Instead of “Take into account” (“berücksichtigen”) you should name it: “also use motorways” or “include motorways”.  Or you could include it in the “avoid” section. “Avoid motorways”. In any case: By default, motorways should be avoided (not included)!

 

3.)   The elevation preference choice is great, and very important. I think it would be good to add it to the quick menu (the menu that you get while routing).  

 

4.)   If you are using the “safe” mode in an urban environment, routing is quite slow for longer routes (25 to 35 kilometers). Maybe this is because elevation is taken into account? Maybe a user switch would be good: “faster (but less exact) calculation”.

 

5.)   It would be great if you could explain on your homepage how elevation data are taken into account.

 

A quick test of the new bicycle routing produced some very good results, but also some results that I did not understand. Will try to test further in the comings months and give you feedback.

 

Thank you and all the best

 

Malte

 

Von: Malte Heynen [mailto:posteinwurf@web.de] 
Gesendet: Dienstag, 22. November 2016 08:07
An: 'osmandapp/OsmAnd-resources'
Betreff: AW: [osmandapp/OsmAnd-resources] Better bicycle routing (#375)

 

1.) Thats right. Maybe we can find another phrase like „more main roads“ / “balanced” / “less main roads”

Or: “include main roads” / “balanced” / “avoid main roads”

 

Something like this.

 

 

 

Von: xmd5a [mailto:notifications@github.com] 
Gesendet: Montag, 21. November 2016 08:30
An: osmandapp/OsmAnd-resources
Cc: Malte; Mention
Betreff: Re: [osmandapp/OsmAnd-resources] Better bicycle routing (#375)

 

@Posteinwurf <https://github.com/Posteinwurf> 

1.	Partially agree, but "Prefer main roads" can be misunderstood as "Prefer ONLY main roads" while this routing prefer cycleways at most.
2)Agree. I'll fix it.

PR #376, opened 2016-04-20
Icons for underground fire hydrant and suction point.

kendzi:
I would like to add icon for tag emergency=suction_point and emergency=fire_hydrant with fire_hydrant:type=underground. It is described on wiki page:
http://wiki.openstreetmap.org/wiki/Tag:emergency%3Dsuction_point
http://wiki.openstreetmap.org/wiki/Tag:emergency%3Dfire_hydrant

I added only icon without any changes to default style. That icon should be optional and used only with custom rendering styles. It is very useful for e.g. for firefighters.

That icon could be accessed by custom style like:

```
<renderingStyle name="fire_hydrant.render" depends="default" defaultColor="#f1eae4" version="1">
<point>
     <filter minzoom="10" icon="emergency_suction_point" tag="emergency" value="suction_point"/>
</point>
</renderingStyle>
```

### DIFF
--- a/icons/svg/emergency/fire_hydrant_underground.svg
+++ b/icons/svg/emergency/fire_hydrant_underground.svg
@@ -1,0 +1,504 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:osb="http://www.openswatchbook.org/uri/2009/osb"
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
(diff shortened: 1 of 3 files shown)


Dzido:
Good idea. We are waiting.

Firefighters from Choszczno

xmd5a2:
We already have fire hydrant icon and it is searchable as POI and even visible at 19+ zooms.
https://github.com/osmandapp/OsmAnd-resources/blob/master/icons/svg/emergency/fire_hydrant.svg

Dzido:
This is  icon for the hydrant:type=underground. In osmand is icon for hydrant:type=pillar.

PR #377, opened 2016-05-02
added more Italian road shields

hakuchi:
I added more Italian road shields.
can someone please fix the different colors from the **ref=[RA](https://it.wikipedia.org/wiki/Raccordo_autostradale#I_raccordi_autostradali_italiani_e_i_gestori)**

**RA4**, **RA6**, **RA7** and **RA16** have blue background, all other **RA** have green background.


xmd5a2:
I think situation with RA should be fixed not in osmand but in OSM. We can't add tons of conditions which are used only for 1 road in the world. Good example is Spain with chaos in road shield colors and road prefixes https://en.wikipedia.org/wiki/Highways_in_Spain#Andalusia

xmd5a2:
We support ref:colour tag ("road_" is internal osmand prefix)
See https://github.com/osmandapp/OsmAnd-resources/blob/master/obf_creation/rendering_types.xml#L1412
https://taginfo.openstreetmap.org/keys/ref%3Acolour

PR #381, opened 2016-06-05
Update es-ar/ttsconfig.p

franco999:
More minor corrections syntax.


franco999:
Thanks for the merger and forgiveness for the many pull request. xD

PR #383, opened 2016-06-28
Map oneway=1 to oneway=yes at obf creation time, not routing time.

aceman444:
Replace oneway=1 with oneway=yes in rendering_types.xml for routing purposes. This should save us the need to evaluate this outdated value at routing runtime (in routing.xml).

### DIFF
--- a/obf_creation/rendering_types.xml
+++ b/obf_creation/rendering_types.xml
@@ -5208,6 +5208,7 @@
 		<routing_type tag="roundabout" mode="amend" base="true"/>
 		<routing_type tag="oneway" mode="amend" base="true"/>
 		<routing_type tag="oneway:bicycle" mode="amend"/>
+		<routing_type tag="oneway" value="1" tag2="oneway" value2="yes" mode="replace"/>
 
 		<routing_type tag="maxspeed" mode="amend" base="true"/>
 		<routing_type tag="maxspeed:practical" mode="amend" base="true"/>
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -177,8 +177,8 @@
 
 		<way attribute="oneway">
 			<select value="1" t="oneway" v="yes"/>
-			<select value="1" t="oneway" v="1"/>
 			<select value="-1" t="oneway" v="-1"/>
+			<select value="0" t="oneway" v="no"/>
 			<select value="1" t="roundabout"/>
 			<select value="1" t="junction" v="roundabout"/>
 		</way>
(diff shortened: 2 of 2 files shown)


vshcherb:
We can't remove <select value="1" t="oneway" v="1"/> cause it is already in many maps. oneway no is not needed cause by default it is 0.

aceman444:
In which maps? In those that people currently have and will get updated (with the new rule in obf creation) until the new OsmAnd with the routing change is released? Or what do you mean?

Yes, 0 is default, but when there is "junction=roundabout+oneway=no" OsmAnd evaluates that as oneway=yes (assuming the first-matching-rule-wins also works in the "oneway" section). I wrote that in the comments. So if you do not like my rule, how do you propose to handle this?

vshcherb:
Yes, "junction=roundabout+oneway=no"  that's possible case. Do we have such data at all?

What I was saying:
<select value="1" t="oneway" v="1"/>
We can't remove this line, cause users might have old maps which doesn't have that preprocessing rule, so it will fail on very important case of detecting oneway road.

aceman444:
Yes, there are many "junction=roundabout+oneway=no" even in Europe (try Overpass turbo), but I am not sure if those are mistakes or are valid.

OK, so if you worry about users on nightlies (which would get the routing change) using old maps, we can first merge the obf change, than later in some months merge the routing change.

vshcherb:
>> OK, so if you worry about users on nightlies (which would get the routing change) using old maps, we can first merge the obf change, than later in some months merge the routing change.
Nope, we should be ready to release every day or at least every month. People could have  a year old maps and it is not our choice and there is no sense to break backward compatibility for this case

aceman444:
I'm not sure a year old map is useful for any routing, but OK, you know your users better.
So there were times when we needed to get new maps for a new OsmAnd version (probably because the rendering data/rules changed). So you can postpone this patch to one of those occasions.

aceman444:
You still haven't fixed the "junction=roundabout+oneway=no" case. So can you merge at least that part?

jfburke3:
I will note that people using the free version might have year-old (or even older!) maps due to the 10-file download limit.  In Europe this might not be an issue, because the maps are very well developed there, but in the U.S., a year is a very big difference in map data.  (I know, because of all of the map edits I've done in that time.)

I am not trying to say that y'all have to genuflect to people who refuse to buy the paid version forever, just pointing out who this would impact.

aceman444:
Thanks. Doesn't your story actually support dropping those users as they do not case about proper routing anyway (using outdated maps)?
I just do not like the situation that we have to punish all users with a runtime performance penalty supporting an old tag, that can be nicely hidden away by the preprocessor.

PR #384, opened 2016-07-09
Make landuse=industrial darken than residential

ad47:
Default OsmAnd style contains different but very close colors for residential and industrial areas. I think industrial should be a bit darker (and of course still no prominent).
https://www.dropbox.com/s/pk7hz3cqstvz2q0/Ind.png?dl=0
https://www.dropbox.com/s/u1h4d8j4nu4rzr9/Ind_new.png?dl=0

### DIFF
--- a/rendering_styles/default.render.xml
+++ b/rendering_styles/default.render.xml
@@ -1127,9 +1127,9 @@
 		<case nightMode="true" attrColorValue="#aaaaaa"/>
 		<case attrColorValue="#666666"/>
 	</renderingAttribute>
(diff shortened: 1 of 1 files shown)


xmd5a2:
Is'nt this color too close to default building color?

ad47:
It does not seem close both to building=yes and industrial
https://www.dropbox.com/s/vete5pr61z6se5h/IndBuildings.png?dl=0

ad47:
Looks like is Europe it used exclusively to mark traffic_signals without pedestrian crossings.

PR #385, opened 2016-07-09
Dont't avoid crossing=no

ad47:
I've found many points marked with crossing=no where route should normally go along highway via this point.
http://www.openstreetmap.org/node/3601443925
http://www.openstreetmap.org/node/2042635591


ad47:
Looks like is Europe it used exclusively to mark traffic_signals without pedestrian crossings.

PR #387, opened 2016-07-11
Consider points with access=private and destination

ad47:
Sometimes highway lines didn't have access tags but access restricted with point barriers. Unfortunately, I found that access private and destination missed in point obstacles. Considering that ways with access=private and destination strongly deprioritized, but not avoided, I've introduced large penalties for this points. I'm not sure about exact values, at least, the longer is the alternative route, the greater they must be, but 60 and 180 seemed to be OK for motorcar (but not for foot, because 3 additional minutes for foot route doesn't change the situation).
Some examples for tests:
1) road between 2 lift gates with access=destination http://map.project-osrm.org/?z=14&center=55.457835%2C36.858187&loc=55.471386%2C36.869988&loc=55.444693%2C36.822696&hl=en&alt=0 (OsmAnd did motorcar route correctly)
2) some gates with access=private http://bit.ly/29JnhVn (motorcar and bike routes are OK, but foot route will lead through private gate until time obstacle less than 360).

Since in <point attribute="obstacle"> section value of 0 means "no penalty" for routing engine I've changed all 1-second penalties to 0.

And I've also removed rest of the <select value="-1" t="crossing" v="no"/> for the reasons described earlier https://github.com/osmandapp/OsmAnd-resources/pull/385

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -377,7 +377,6 @@
 			<select value="0"  t="motorcar" v="yes"/>
 			<select value="0"  t="motorcar" v="permissive"/>
 			<select value="0"  t="motorcar" v="designated"/>
-			<select value="0"  t="motorcar" v="destination"/>
 			<select value="0"  t="motorcar" v="customers"/>
 			<select value="0"  t="motorcar" v="official"/>
 			
@@ -389,7 +388,6 @@
 			<select value="0"  t="motor_vehicle" v="yes"/>
 			<select value="0"  t="motor_vehicle" v="permissive"/>
 			<select value="0"  t="motor_vehicle" v="designated"/>
-			<select value="0"  t="motor_vehicle" v="destination"/>
 			<select value="0"  t="motor_vehicle" v="customers"/>
 			<select value="0"  t="motor_vehicle" v="official"/>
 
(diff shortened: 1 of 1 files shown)


routing/routing.xml
```diff
@@ -372,39 +372,47 @@
...
 			<select value="-1" t="motorcar" v="agricultural"/>
 			<select value="-1" t="motorcar" v="forestry"/>
-			<select value="1"  t="motorcar" v="yes"/>
-			<select value="1"  t="motorcar" v="permissive"/>
-			<select value="1"  t="motorcar" v="designated"/>
-			<select value="1"  t="motorcar" v="destination"/>
-			<select value="1"  t="motorcar" v="customers"/>
-			<select value="1"  t="motorcar" v="official"/>
+			<select value="180"  t="motorcar" v="private"/>
+			<select value="60"  t="motorcar" v="destination"/>
```
xmd5a2: motorcar=destination already exists at line 380

ad47:
Yes, I overlooked this lines and removed them now. I also reverted penalties for foot to -1, since I have no idea, how large they must be to avoid all possible routes via access=private points.

PR #388, opened 2016-07-12
Allow cyclists to use living_streets

ad47:
I dotn't want to discuss, should osmand strictly obey rules or not,  but set bicycle=destination on highway=living_street is really not a good idea. Alternative route will probably go via alongside footways, highway=service+living_street=yes or via main roads.

Here is an example http://bit.ly/29zjXqk (inconvenient route with 2 left turns, but osmand did something weird https://www.dropbox.com/s/cqtjcmltvkbnxei/weird_route.png?dl=0).


aceman444:
Did you notice that rule only applies to Russia and Ukraine? Do you know the laws there? That rule was only put there recently, there surely was a reason.

PR #391, opened 2016-07-17
Add bicycle=official

ad47:
Paths with bicycle=official should be rendered the same way as ones with bicycle=designated.

### DIFF
--- a/obf_creation/rendering_types.xml
+++ b/obf_creation/rendering_types.xml
@@ -138,6 +138,7 @@
 	<type tag="bicycle" value="permissive" minzoom="13" additional="true" poi="false"/>
 	<type tag="bicycle" value="destination" minzoom="13" additional="true" poi="false"/>
 	<type tag="bicycle" value="designated" minzoom="12" additional="true" poi="false"/>
+	<entity_convert pattern="tag_transform" from_tag="bicycle" from_value="official" to_tag1="bicycle" to_value1="designated"/>
 	<type tag="fee" value="yes" minzoom="13" additional="true" poi="false"/>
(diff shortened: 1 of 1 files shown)


aceman444:
Will this replace all bicycle=official with bicycle=designated also for routing purposes? You can see we handle both values in routing.xml (and not identically).

ad47:
Yes, it will. I can not see any difference between official and designated on the wiki page (except, official is "stronger"), so should it be any difference in routing?

xmd5a2:
@aceman444 Fixed https://github.com/osmandapp/OsmAnd-resources/commit/07854e9688b51b00b93c3c6cbae3070760b827ad

xmd5a2:
@ad47 "official" has more priority in routing.xml [https://github.com/osmandapp/OsmAnd-resources/blob/master/routing/routing.xml#L594](url)
Let's keep it.

ad47:
Ok, but it looks strange, since difference between this tags it pretty unclear.

aceman444:
The rendering can be the same of course. Yes, 'official' implies to be stronger (designated by state/road owner) with a sign, while 'designated' can be without sign, just that the random mapper thinks this road seems like best for bikes.

ad47:
designated should not be be without sign, bicycle=yes is for it :)
So you really think there is a difference between 2 sides of the street?
http://www.openstreetmap.org/way/371065764
http://www.openstreetmap.org/way/121899537

aceman444:
Sure, one may have a sign :)
I haven't invented the tags/values but those seem to be definitions :)
On the other hand I do not understand why bicycle_road=yes is needed (seems to be the same as highway=* + access=no + bicycle=official), but it seems some people have a use for it.

PR #392, opened 2016-07-18
Add bicycle_road=yes

ad47:
http://wiki.openstreetmap.org/wiki/Key:bicycle road
Bicycle roads are much better for cyclists than common roads, so it worth to prefer them in routing.xml and render additional dashed line similar to cycleway=lane, but using cycleway style.
https://www.dropbox.com/s/bh9tk47touiz6e9/bicycle_road.png?dl=0

### DIFF
--- a/rendering_styles/default.render.xml
+++ b/rendering_styles/default.render.xml
@@ -2181,6 +2181,11 @@
 					<apply_if appMode="default" moreDetailed="false" maxzoom="15" order="-1"/>
 				</apply_if>
 			</case>
+			<case appMode="bicycle" tag="bicycle_road" value="yes" order="143">
+				<apply_if layer="1" order="144">
+					<apply_if appMode="default" moreDetailed="false" maxzoom="15" order="-1"/>
+				</apply_if>
+			</case>
 			<switch>
 				<switch>
 					<case tag="highway" value="path" order="37">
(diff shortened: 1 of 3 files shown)


rendering_styles/default.render.xml
```diff
@@ -2181,6 +2181,11 @@
 					<apply_if appMode="default" moreDetailed="false" maxzoom="15" order="-1"/>
 				</apply_if>
 			</case>
+			<case appMode="bicycle" tag="bicycle_road" value="yes" order="143">
+				<apply_if layer="1" order="144">
+					<apply_if appMode="default" moreDetailed="false" maxzoom="15" order="-1"/>
```
xmd5a2: This condition will never happen because it is inside of appMode="bicycle"

ad47:
Sounds logical, so I removed both redundant lines.

aceman444:
What is this patch about? I see no bicycle_road key on that linked wiki page.
Also that key is never used in the database: http://taginfo.openstreetmap.org/search?q=bicycle_way#keys 

There must be some mistake here. Please revert this until it is resolved.

xmd5a2:
Bicycle "road", not "way". There is a space in a link.

aceman444:
I see, the link went to Key:bicycle ;)

PR #393, opened 2016-07-21
Update poi_types.xml

hvdwolf:
Add additional user defined tag for UK postcode POI files

### DIFF
--- a/poi/poi_types.xml
+++ b/poi/poi_types.xml
@@ -3024,6 +3024,8 @@
 			</poi_type>
 		</poi_category>
 		<poi_category name="user_defined_other" default_tag="amenity" no_edit="true">
+			<!-- Add additional combination for UK postcode POI files-->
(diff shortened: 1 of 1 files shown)


hvdwolf:
I did a pull request for the poi_types.xml. 
Like mentioned in the [Osmand-tools issue 128](https://github.com/osmandapp/OsmAnd-tools/issues/128), I did add my own user defined POI type directly to the jar, and that works fine as such.
The POI file can be used in side OsmAnd and usr can search on it. It can also be displayed as an "orange dot" in the map, but when clicking on that POI and requesting details (the 3 vertical dots in the bottom-right), Osmand crashes as the apk doesn't know that POI type.
So it needs to be added to the apk as well.

And maybe it needs some error checking as well in the apk that Osmand doesn't crash when encountering an "incorrect" POI.

PR #394, opened 2016-08-03
Rework of the bicycle routing

ad47:
Brief description:
Main advantage - paths and tracks with surface=asphalt/paved/etc. has the same priority as cycleways and roads with bicycle lanes (the idea comes from [this](https://github.com/osmandapp/OsmAnd-resources/pull/375) PR).

Roads with bicycle lanes and cycleways has priority 1.2 that means they will be preferred even if the route becomes up to 20% longer (it was enough in my tests).

Option "Prefer minor roads" deprioritize major highways and route will go via cycleways, small streets or footways (only if there are no other options). I've also found some interesting "off-road" routes outside the city using this option.

I've attached tables with speeds, (first matching column is applied). It still looks like road-bike profile, but more balanced in my opinion. I would be glad if some could test it and compare with current profile.
[Bicycle_routing_speeds.xlsx](https://github.com/osmandapp/OsmAnd-resources/files/400219/Bicycle_routing_speeds.xlsx)

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -698,7 +698,7 @@
 			<select value="0"  t="access" v="yes"/>
 			<select value="0"  t="access" v="permissive"/>
 	
-			<select value="15" t="highway" v="traffic_signals"/>
+			<select value="10" t="highway" v="traffic_signals"/>
 			<select value="10" t="crossing" v="traffic_signals"/>
 			<select value="10" t="barrier" v="cycle_barrier"/>
 			<select value="10" t="barrier" v="debris"/>
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -437,6 +437,7 @@
 		<!-- New ROUTING API -->
 		<parameter id="avoid_ferries" name="Avoid ferries" description="Avoid ferries" type="boolean"/>
 		<parameter id="avoid_motorway" name="Avoid motorways" description="Avoid motorways" type="boolean"/>
+		<parameter id="prefer_minor_roads" name="Prefer minor roads" description="Prefer small streets and cycleways" type="boolean"/>
 		<parameter id="avoid_unpaved" name="Avoid unpaved roads" description="Avoid unpaved roads" type="boolean"/>
 		<parameter id="avoid_stairs" name="Avoid stairs" description="Avoid stairs" type="boolean"/>
 		<parameter id="avoid_borders" name="Avoid border crossing" description="Avoid crossing a border into another country" type="boolean"/>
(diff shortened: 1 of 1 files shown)


xmd5a2:
Please note that it is much easier to check pull requests with minimum changes. Large pull requests are less likely to be merged.

ad47:
I know and I have no hope it will be merged soon. Unfortunately, many parameters affect the others, so it was easier to test only the final result. Anyway, I can comment every line or each block if it necessary.

vshcherb:
Unfortunately I have no much right to merge it, cause I was never responsible for bicycle routing and don't know much about except some basics. I think @aceman444 and others contributed much to bicycle routing.

As I see some concepts are changed much.

vshcherb:
From my end I could only point some changes that most likely dangerous:
1. Reducing max speed from 27 to 24. It could has dramatic effect on the route calculation time as well as deprioritizing cycleway. I think now all routes of 100-250km will be calculated much slower in dense area such as Benelux.
2. <if param="prefer_minor_roads"> changing maxspeed is a bad practice, there is a priority section for that.  It works exactly the same and do not mix 2 concepts of reasonable speed on the way and on priority to select. So saying motorway=2kmh is definitely unreasonable cause pedestrian would move faster.

3. 
<select value="1.15" t="cycleway" v="share_busway"/>
<select value="1.5" t="" v="grade1"/>
I guess you know that priorities are not multipled. So if you have (cycleway=share_busway, tracktype=grade1), priority will be 1.15 and not 1.5*1.15. I'm afraid deprioritising cycleway from 1.6 is a conceptual mistake. And it could influence on the calculation speed

ad47:
Thank you for your comment, Victor.

1. I didn't know, that _maxDefaultSpeed_ has an influence on route calculation time. I still can't understand its effect on calculated route, but later I will set it to 26-28.

2. I've used way speeds in "prefer minor roads" because only first priority is applied and if there is cycleway tag on "unwanted" highway, it will left with the same low priority. I've changed this section using priorities.
 
3. About priority 1.6 for cycleways - it already restricted by _maxDefaultSpeed_, so it actually 1.5. 
But I've found coefficient 1.2 is sufficient for adequate routes. I agree, 1.3-1.4 might be better, but not 1.6 - who wants 60%-longer routes, even via cycleways? Moreover, when I tried extremely large priority for cycleways with large _maxDefaultSpeed_, route calculation speed in Moscow was dropped dramatically, so may be smaller speed range is better? 
There is also one side effect of large priority - sometimes Osmand tries to include cycleway in route by any price. 

cycleway=share_busway is rather is a bus lane, were cyclists are allowed, it better, than common highway, but not as cycleway/lane, so I set a bit smaller priority.

vshcherb:
I'm closing this issue/pull request cause we have no power to support custom routing yet. We will be working on unit testing of bicycle routing. In parallel I would like to request to keep all your findings. Our first plan is to introduce "author view" routing, very similar concept to rendering style. So end user could choose offline routing.xml based on description of routing.xml.  https://github.com/osmandapp/Osmand/issues/3188

PR #396, opened 2016-08-09
Update config.p for Slovak

aceman444:
This file was generated from the sk/ttsconfig.p via gen_voice.sh.

I also gave the generated ogg files, that were created using 'gen_voice.sh sk Slovak google sk' and TEMP_FACTOR=1.3 . Should I upload it here somehow? If yes, how? Or will you generate the package yourself?

This package replaces sk-Ruzena package, that was outdated and is no longer offered for download.

### DIFF
--- a/voice/sk/config.p
+++ b/voice/sk/config.p
@@ -1,5 +1,5 @@
 %%% !!! THIS IS GENERATED FILE !!! Modify ttsconfig.p
-﻿% for turbo-prolog
+% for turbo-prolog
 :- op('--', xfy, 500).
 % for swi-prolog
 :- op(500, xfy,'--').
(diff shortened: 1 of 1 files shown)


vshcherb:
Why would we update config.P if we don't update voice files? config.P is usually generated from ttsconfig.P but it should be committed always with voice files, otherwise it just useless and breaks current installation

vshcherb:
Of course I could be mistaken, if there are new features that didn't require new files and other new features are ok to be muted

xmd5a2:
config.p will be overwritten after gen_voice.sh start

aceman444:
vshcherb, I asked if I should upload the voice files too and how. So yes, new voice files are needed for this config.p but I ask if I should upload them somewhere or you will generate the automatically. I uploaded ttsconfig.p recently, that can be used to create this config.p

xmd5a2, I see no "start" argument to gen_voice.sh. If you mean gen_all.sh, then there is no rule for Slovak in it. You could add "gen_voice.sh sk Slovak google sk" which I wrote. But I am not sure where you would write the tempo rate.

vshcherb:
We usually use TTS voices cause then we can regenerate it once new features will be added. @xmd5a2  working on the script to use Google TTS.

aceman444:
No problem with that. The master ttsconfig.p for this language was merged in commit dcf082445704fe5870b882358c2caa003e792f7d .
I'm fine if you generate the pre-recorded voices from it. There just need to be 
1. generate the voices via 'gen_voice.sh sk Slovak google sk' with TEMP_FACTOR=1.3.
2. expose the new 'sk' package for download (as sk-Ruzena is not exposed) through OsmAnd.
3. then optionally remove sk-Ruzena at the right time according to the purging policy.

Please make the new 'sk' prerecorded voices available to download in OsmAnd 2.4 and up.

Thanks.

xmd5a2:
@aceman444 done

aceman444:
I see you have generated the new ogg files for Slovak. The ogg file sound fine.
Unfortunately, some old config.p file went into the package so the voice commands are quite broken. The file is 13Kb, while the one we have commited here is 19KB. Please fix that.

aceman444:
Strangely, the package at http://download.osmand.net/rawindexes/list.php?sortby=cmp_date&d=-1 contains the correct _config.p. May there be a problem (out of sync) on the download servers?

aceman444:
Ok, when downloading from inside OsmAnd 2.4, the proper _config.p is downloaded in the package. So probably the only thing to do here is to not offer the sk pre-recorded package to OsmAnd 2.3 users as it is just broken.

PR #400, opened 2016-10-26
fr-tts: Fix and improve French words

mathieumd:
Improve French translation by using meaningful words and actually used sentences (ie. not Google-translated ones!).


sonora:
Thanks!

PR #406, opened 2016-11-21
Some fix in arabic tts file

BoFFire:
Some fix in arabic tts file.


vshcherb:
Is it going to merge? 
CC @sonora

sonora:
I was confused, because it is closed. But I see no reason why we should not merge it?

sonora:
Looks like we're good, it's all in. Thanks, @BoFFire !

BoFFire:
Sorry. My bad. I'm not used with Github but the changes are tested on Vocalizer TTS. Working fine. Some more tts prompts on this file to correct and everything will be OK.

sonora:
Yes, I am not the best Github expert either :)

Ok, I have just merged what I have found so far. It will be in our nightly build in a few minutes, so you can download the apk test, if you want. Please add more pull request, if you can.

I am surprised we need no corrections to grammar or word order?

BoFFire:
Thank you @sonora Ok I'll update my OSMAnd app to test the changes. Ok for the pull requests :)
For now, the word order seems okay. Don't be surprised. May be one word but spelled well in general. Yep I add some more corrections tomorrow.
Thank you all.

PR #411, opened 2016-12-11
Add surface=earth, same color as dirt, see osm wiki surface

it4workflow:


xmd5a2:
surface=earth already supported
https://github.com/osmandapp/OsmAnd-resources/blob/master/obf_creation/rendering_types.xml#L2054

it4workflow:
WTF

PR #412, opened 2016-12-12
fix spelling / pronounciation of "Straßenverlauf"

buxit:
Strasenverlauf -> Straßenverlauf


buxit:
is this the only place where this needs to be fixed?

hakuchi:
if you change the "s" to "ß", some TTS voices can have problems to use this correct.

sonora:
See https://github.com/osmandapp/OsmAnd-resources/pull/414

PR #413, opened 2016-12-12
fix "Straßenverlauf" spelling in ttsconfig.p

buxit:
Strasenverlauf -> Straßenverlauf

### DIFF
--- a/voice/de/ttsconfig.p
+++ b/voice/de/ttsconfig.p
@@ -79,7 +79,7 @@ string('17th.ogg', 'siebzehnte ').
 
 % STRAIGHT/FOLLOW
 string('go_ahead.ogg', 'Weiter geradeaus ').
-string('follow1.ogg', 'Dem Strasenverlauf ').
+string('follow1.ogg', 'Dem Straßenverlauf ').
 string('follow2.ogg', 'folgen ').
 
 % ARRIVE


hakuchi:
if you change the "s" to "ß", some TTS voices have problems to use this correct.

hakuchi:
If you will use this for German-TTS the only file you have to change is this.
The other two request you can delete, because they fixing/changing nothing in TTS.

I tested this with some different TTS engines and the first say "Dem Strßenverlauf" and Google TTS say "dem Stassenverlauf". On streets the name(example "Bahnhofstraße") sounds good.

If I use the `ttsconfig.p` without any changes, the two TTS engines sounds always correct on street name and follow street for xx kilometers.

You can test this by yourself when you open `ttsconfig.p` in your osmand folder on your phone.

I think this should not be changed!

buxit:
besides the tts i also want to fix the spelling everywhere, since Strasenverlauf is not correct.

i just tested this (modifying ttsconfig.p) on two different devices and it works perfectly. i still suspect your encoding is wrong. bad unicode character or similar?

hakuchi:
The "Strasenverlauf" inside `ttsconfig.p` is only used for the TTS voice and for nothing other!
If in OsmAnd the name from the street is "Bahnhofstraße" or "Bahnhofstrasse", Osmand using the name from the OSM database.

As I writhed above:
**Some devices will having problems if you change the "s" to "ß"**

PS: And if next time comes a user from Switzerland we have to change from your "Straßenverlauf" to "Strassenverlauf"?

PPS: If you will have "Straßenverlauf" in your `ttsconfig.p` feel free to change this in your osmand folder.

vshcherb:
I would say Strassenverlauf would be better, though let's wait the reply from @sonora

sonora:
Well, this is not a spelling contest ... :)

@vshcherb I did actually put "Strasenverlauf" on purpose because (way back when) that produced the best pronunciation on the TTS engines I had tested then (I think this was several years ago, and what was tested were the then current versions of PICO TTS, Google TTs, and one commercial one, the name of which I cannot remember).

Putting an "ss" resulted in a short syllable pronunciation, which sounded really weird and deserved changing. And some engines, although they claimed supporting German, actually had severe pronunciation issues with "ß".

While the engines have certainly developed, and things may have improved, I think we should not change this
- unless we have a real pronunciation issue as is (which I cannot see on any of my devices)
- have tested (you can produce your own ttsconfig.p to test) that a change would be a real improvement.

PR #414, opened 2016-12-12
fix spelling / pronounciation of "Straßenverlauf"

buxit:


buxit:
sorry, all of these should have been changed in one commit

hakuchi:
if you change the "s" to "ß", some TTS voices can have problems to use this correct.

buxit:
1. the pronounciation is really strange when this is spelled with an "s".
1. _"Dem Strasenverlauf folgen"_ also appears as textual information — this is bad spelling.
1. many streets are named *straße, so people with bad TTS versions will be used to the problems when the word Straße is synthesized.
1. possibly the wrong symbol was used when you first tested this?
1. i opt for trying again with the correct spelling or changing the phrase altogether

sonora:
Ok, let's try "Strassenverlauf" and see what feedback we draw when this is tested by more people on more devices. (https://github.com/osmandapp/OsmAnd-resources/commit/2453bae4c9b4a9ff60fbc6879acaf4b1333c9c3d)

From my testing, "Stra0enverlauf" would produce no different result in German for Germany and Austria, but is a no go because it may potentially break things for Swiss German engine settings (I cannot fully test, so no point in running the risk).)

buxit:
looks good to me. according to my tests the pronunciation is the same on recent android versions.

sonora:
Ok, agreed, I think we can leave it as it is now! Thanks!

PR #418, opened 2017-01-01
Little grammar correction

mircozorzo:


sonora:
This is a generated file, it needs to be fixed in ttsconfig.p (I just did).

mircozorzo:
Thanks so much.  mercoledì, 04 gennaio 2017, 06:54PM +0100 da Hardy  notifications@github.com :

PR #421, opened 2017-01-05
Arabic TTS updates

BoFFire:
Arabic TTS updates


vshcherb:
Could not be merged because of conflicts

PR #423, opened 2017-01-22
rendering: optional cycle route highlighting

exaexa:
Default highlighting of cycle routes is sometimes too narrow to be seen,
sometimes just recoloring the road to something that looks like higher-traffic
road etc. Add an option to make the highlight 2x wider to be easier to spot.
This also makes it look a bit like OpenCycleMap route highlighting, which I
consider to be a pretty good guideline.

On the picture (with older default.render.xml, where the situation was even worse):
- left: no cycle routes
- mid: cycle route highlighting makes it hard to guess whether it's a cycle route, or a different road type
- right: 2x wider highlighting, everything's clear on a first look.

osmand

### DIFF
--- a/rendering_styles/default.render.xml
+++ b/rendering_styles/default.render.xml
@@ -92,6 +92,8 @@
 	<renderingProperty attr="contourLines" name="Show contour lines" description="Select minimum zoom level to display in map if available. Separate contour file may be needed."
 		type="string" possibleValues="disabled,16,15,14,13,12,11" defaultValueDescription="default13" category="details"/>
 	<renderingProperty attr="contourColorScheme" name="Contour lines color scheme" description="Contour lines color scheme" type="string" possibleValues="light_brown,brown,dark_brown" defaultValueDescription="default"/>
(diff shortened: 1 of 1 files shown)


marshall2389:
I'm hoping to adjust my renderer to make cycle paths and footpaths more visible. Could you help me? I don't know how to edit the depends-template to get the results I'm looking for.

Thanks,
Elliot

xmd5a2:
No option is needed. Merged manually.
https://github.com/osmandapp/OsmAnd-resources/commit/a681f41e5089b4ce8cdd27e6e1e575fab61141cf

exaexa:
Thank you! :+1:

exaexa:
Hello,
It seems that this issue is "back" after commit 37578133aaca9c329706fdc3d2aa1c4e28e12cb7 where @xmd5a2 changed the bike route width back. Please refer to the original images to see the problem with bike route highlighting.

Is there any reason for reverting this change?

If it breaks some other rendering, perhaps a conditional code or user switch would work? (I can submit the code).

Thanks in advance!
-mk

PR #426, opened 2017-02-16
Updating Arabic TTS

BoFFire:
Updating Arabic TTS and removing some spaces.


BoFFire:
Updating Arabic TTS and removing some spaces.

PR #428, opened 2017-02-28
Do not ignore barriers when choosing shortest way.

aceman444:
With barriers not listed in the routing.xml file, we must get to the rule <select value="-1" t="barrier"/> even for shortest way, it is crucial for the "obstacle" section to work properly.
This should fix issue #3095, which is a serious problem.
It will probably regress #1778, which will need a different solution.

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -418,9 +418,6 @@
 			<select value="0"  t="access" v="yes"/>
 			<select value="0"  t="access" v="permissive"/>
 			<select value="0"  t="access" v="customers"/>
-			<if param="short_way">
-				<select value="0"/>
-			</if>
 
 			<select value="5"   t="barrier" v="cattle_grid"/>
 			<select value="300" t="barrier" v="border_control"/>


maraf24:
You can retain the logic with this code:
```
<if param="short_way">
  <select value="0"   t="barrier" v="cattle_grid"/>
  <select value="0" t="barrier" v="border_control"/>
  <select value="0" t="barrier" v="bump_gate"/>
  <select value="0"   t="barrier" v="entrance"/>
  <select value="0" t="barrier" v="height_restrictor"/>
  <select value="0" t="barrier" v="sally_port"/>
  <select value="0"   t="barrier" v="toll_booth"/>
  <select value="0" t="barrier" v="lift_gate"/>
  <select value="0" t="barrier" v="swing_gate"/>
  <select value="-1" t="barrier"/>         
  <select value="0"/>
</if>
```

aceman444:
It is unfortunate to duplicate all those tags, but yes, that could also work. Can you make the patch?

xmd5a2:
Merged manually https://github.com/osmandapp/OsmAnd-resources/commit/a9f2ba48bd556681c46a64b7521537911b333525

aceman444:
Thanks. This should go into 2.6.x as it creates very dangerous and invalid routes for cars.

PR #433, opened 2017-05-19
Proposals for bicycle routing - Part 1

Posteinwurf:
Hello to everyone who is involved with Osmand development,

thank you for the new bicycle routing! The inclusion of elevation data is incredibly good. I am really impressed by how you implemented this! The graphic display was the biggest surprise - and works perfectly. 

I have some suggestions for further improvements in bicycle routing. This is the first part (second will follow in the next days). My two suggestions are below in the commentary (in the routing.xml I have just deleted an " ß " . This character leads to some android browsers being unable to display the file). 

Now, my first two suggestions:

1.)  If a footway is marked as “sidewalk”, this is not recognisable in the Osmand maps at the moment. There is already a corresponding command in the routing.xml, but at the moment, it has no effect: _select value="0.3" t="footway" v="sidewalk"_ To make this command working, we have to include the “footway = sidewalk” data into the Osmand maps. Could someone implement this? (I do not know how to do it!). This would help to avoid many routing errors (for example in Vienna, where I have tested the new routing last week).

2.) There is still one bug concerning traffic signals. When a bicycle is running on a main road and approaches a big junction, it will often "hit upon" two tags for the traffic signals: First, there is a tag "highway=traffic_signal" (the traffic light for cars / bicycles on the road). Then, right behind, there is a tag "crossing=traffic_signals" (the traffic light for pedestrians crossing - sometimes this is just marked as "crossing=island", but not always). In Osmand, both tags have a "penalty" measured in seconds (the penalty is defined under _point attribute="obstacle"_). BUT: When the bicycle is running on the main road, Osmand adds up both penalties, which is wrong: There should be only one penalty, because you do not have to wait two times when you are on the road (you just wait at the car traffic light, after it turns green, you pass the pedestrian traffic light automatically). 

Again, unfortunately, I cannot implement the solution myself. (Suggested solution - please see below, nin the next comment). 

I think it would be important to fix both issues, because they lead to some really weird errors in routing calculation.

Thank you!

Malte

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -1067,12 +1067,11 @@
 			<select value="0.05" t="foot" v="private"/>
 			<select value="0.05" t="foot" v="destination"/>
 
-			<!-- Q? access=destination should not be used foot or what is the proposed schemeß
+			<!-- Q? access=destination should not be used foot or what is the proposed scheme
 			http://www.openstreetmap.org/way/173188624#map=18/39.56805/2.64846
 			should be 
 			<select value="0.05" t="access" v="private"/>
-			<select value="0.05" t="access" v="destination"/>
-		-->
+			<select value="0.05" t="access" v="destination"/> -->
 
 			<select value="1.2" t="sidewalk" v="yes"/>
 			<select value="0.9" t="sidewalk" v="no"/>


Posteinwurf:
I have done additional tests: Actually, at many big junctions, Osmand will impose THREE penalties for a single traffic light. Of course, this leads to severe routing errors in bicycle mode.

The picture below shows the typical mapping of a bigger junction. The red arrows are the direction of travel. The yellow bubbles indicate penalties in seconds that will be applied (three for a single junction). If we set a penalty of 33 seconds for "highway=traffic_signals" and 25 seconds for "crossing=traffic_signals", then this junction will have a penalty of 83 seconds. But it should be only 25 OR 33 seconds for the whole junction.

crossings

Could someone fix this, please? A great solution would be (I hope it is possible): If there is a pedestrian traffic light (marked with "crossing=traffic_signals"), then the penalty should only apply to the pedestrians / bicycles that are CROSSING the road, but not to the bicycles that are travelling ALONG the road. That solution would be almost perfect. 

A correct calculation of penalties for traffic lights at junctions is very important for all bicycle modes. For the "balanced mode" it is important to put in a value that reflects real average waiting times at traffic lights. Then, Osmand will prefer routes without traffic lights (even if they are slightly longer, they will still be faster!).  

For the third bicycle mode (the one that avoids main roads), the penalty for traffic lights is even more important: Here, we need a value that is a bit higher than the average waiting time. This higher penalty is the one main factor that actually helps stay away from main roads. (For various reasons, you cannot achieve this with the values for primary / secondary / tertiary roads alone. You need the slightly higher penalty for traffic lights additionally - I have done extensive testing on this. If someone wants to know more, I can give more details). 

Thanks to all

Malte

vshcherb:
``` the penalty should only apply to those pedestrians / bicycles that are CROSSING the road, but not to the bicycles that are travelling ALONG the road```
Not sure how possible it is to implement.

vshcherb:
In my understanding the tagging is not complete, it should be traffic_signals:forward, traffic_signals:backward. Though I'm not sure if our routing engine could support this.

Posteinwurf:
Unfortunately, in the OSM data, the direction of the traffic lights is not indicated in most cases. https://www.openstreetmap.org/node/677136803 
https://www.openstreetmap.org/node/3411532520
But maybe that could be implemented when the Osmand Maps are generated?

Another solution that might be possible: If a highway=traffic_signals is followed by a crossing=traffic_signals within a very short distance, then the highway=traffic_signals could just be ignored (no penalty). This solution would be a compromise, but would still work reasonably. 

But the best compromise would still be to ignore all crossing=traffic_signals tags, if the bicycles is on the main road.

PR #434, opened 2017-05-24
Fixes and further improvements for bike routing

Posteinwurf:
Hello to everyone,

thanks for the much improved bike routing (including the great and well done inclusion of height data). 

I have done much testing of the latest version (in Vienna, in a hilly countryside, in Berlin, in Salzburg, in Munich). Here are some further improvements for the routing.xml that will minimize errors in bike routing .

Some problems that still exist cannot be fixed within the routing.xml alone. These are:

1.) Osmand still routes through elevators, for example in Berlin (highway=elevator has to be recognised by the routing engine, so that the -1 command can work). 

2.) The routing engine does not recognise footway=sidewalk. It is important that Osmand does include this, so that footway=sidewalk can be heavily downgraded (ca. 0.5) . Without downgrading, this leads to many routing errors (for example in Vienna and Berlin).

3.) At many bigger junctions, there is a triple counting of traffic lights (I have explained the details in my former post).

4.) The height data worked great in Austria, but in Berlin (maybe in other cities also?), the height dataset seems to be incorrect. When you calculate a route that has some hills in the middle, the height profile just shows a flat line in the middle. At the same time, you get a random elevation of some meters at the end (where actually there is no elevation).

Thanks to all,

Malte

P.S. Is there any text explaining how the new osmand_highway_integrity_brouting is calculated? Or could someone give me a brief explanation?

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -484,7 +484,7 @@
 	</routingProfile>
 
 	<routingProfile name="bicycle" baseProfile="bicycle" restrictionsAware="true" minDefaultSpeed="2" maxDefaultSpeed="37"
-			leftTurn="0" rightTurn="0" followSpeedLimitations="false" onewayAware="true" heuristicCoefficient="1.4">
+			leftTurn="5" rightTurn="3" followSpeedLimitations="false" onewayAware="true" heuristicCoefficient="1.4">
 		<!-- <attribute name="relaxNodesIfStartDistSmallCoeff" value="2.5"/>  -->
 		<!-- New ROUTING API -->
 
@@ -499,13 +499,14 @@
 		<parameter id="relief_smoothness_factor_plains" group="relief_smoothness_factor" name="Plains" description="" type="boolean" default="true"/>
 		<parameter id="relief_smoothness_factor_hills" group="relief_smoothness_factor" name="Hills" description="" type="boolean"/>
 
-<!--		<parameter id="prefer_main_highways" name="Prefer main highways" description="Prefer main highways" type="boolean"/>
-		<parameter id="avoid_traffic_signals" name="Less Traffic Signals - Less Main Roads" description="Avoid Traffic Signals" type="boolean"/>-->
 		<parameter id="avoid_unpaved" name="Avoid unpaved roads" description="Avoid unpaved roads" type="boolean"/>
 		<parameter id="avoid_ferries" name="Avoid ferries" description="Avoid ferries" type="boolean"/>
 		<parameter id="avoid_stairs" name="Avoid stairs" description="Avoid stairs" type="boolean"/>
 		<parameter id="avoid_borders" name="Avoid border crossing" description="Avoid crossing a border into another country" type="boolean"/>
 		<parameter id="allow_motorway" name="Allow motorways" description="Allow motorways" type="boolean"/>
+		<!-- line above: is there a possibility to set "allow_motorway" OFF by default? In most
+		countries it is absolutely forbidden to use them with a bicycle. So we should
+		make sure that the user can only turn this option on ACTIVELY. Maybe even a pop up warning? --> 
 		<parameter id="height_obstacles" name="Use elevation data" description="Use terrain elevation data provided by SRTM, ASTER and EU-DEM" type="boolean"/>
 
 		<way attribute="access">
@@ -529,7 +530,18 @@
 				<if param="avoid_stairs"/>
 			</select>
 
+			<select value="-1" t="highway" v="motorway"/>
+			<select value="-1" t="highway" v="motorway_link"/>
+			<select value="-1" t="highway" v="trunk"/>
+			<select value="-1" t="highway" v="trunk_link"/>
+			<!-- lines above: just moved these lines up here
+			to make sure that motorways are excluded if they
+			are not explicitly allowed by the user.-->
 			<select value="-1" t="highway" v="elevator"/>
+			<!-- line above: The Osmand routing engine still does
+			not recognise elevators. So the -1 does not work here
+			(e.g. the routing DOES use elevators at the moment!)
+			Could someone fix this? Important, causes some errors. --> 
 			<select value="-1" t="highway" v="construction"/>
 			<select value="-1" t="construction" v="yes"/>
 			<select value="-1" t="motorroad" v="yes"/>
@@ -549,7 +561,15 @@
 			<select value="-1" t="mtb:scale" v="5"/>
 			<select value="-1" t="mtb:scale" v="6"/>
 
-			<select value="-1" t="vehicle" v="no"/>
+			<select value="1" t="vehicle" v="no"/>
+			<!-- The tag vehicle=no is normally used when mappers
+			want to exclude motor vehicles. I have never seen a case
+			when such a road was actually forbidden for bicycles as
+			well. (Mappers often do not bother to add the bicycle=yes).
+			Example:www.openstreetmap.org/way/159679495  This is
+			even part of a bike route, but nevertheless only marked
+			vehicle=no. If a way is forbidden for bicycles,
+			there will be an "acces=no" instead of "vehicle=no".-->
 			<select value="1" t="vehicle" v="agricultural"/>
 			<select value="1" t="vehicle" v="forestry"/>
 			<select value="1"  t="vehicle" v="yes"/>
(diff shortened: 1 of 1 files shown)


routing/routing.xml
```diff
@@ -953,7 +1055,7 @@
 
 		</point>
 
-		<way attribute="penalty_transition">
+		<!-- <way attribute="penalty_transition">
```
xmd5a2: It was added because of frequently routing engine attempts to avoid main highway which starts with left turn. Left turn on main highways is very dangerous for bicycle and in some countries it is not allowed at all. So I disagree with this change.

xmd5a2:
brouting was introduced because regular osmand_highway_integrity has a slightly different purpose (rendering but not routing). https://github.com/osmandapp/OsmAnd-tools/blob/master/OsmAndMapCreator/src/net/osmand/osm/MapRenderingTypesEncoder.java#L1082

routing/routing.xml
```diff
@@ -499,13 +499,14 @@
...
-<!--		<parameter id="prefer_main_highways" name="Prefer main highways" description="Prefer main highways" type="boolean"/>
-		<parameter id="avoid_traffic_signals" name="Less Traffic Signals - Less Main Roads" description="Avoid Traffic Signals" type="boolean"/>-->
 		<parameter id="avoid_unpaved" name="Avoid unpaved roads" description="Avoid unpaved roads" type="boolean"/>
 		<parameter id="avoid_ferries" name="Avoid ferries" description="Avoid ferries" type="boolean"/>
 		<parameter id="avoid_stairs" name="Avoid stairs" description="Avoid stairs" type="boolean"/>
 		<parameter id="avoid_borders" name="Avoid border crossing" description="Avoid crossing a border into another country" type="boolean"/>
 		<parameter id="allow_motorway" name="Allow motorways" description="Allow motorways" type="boolean"/>
+		<!-- line above: is there a possibility to set "allow_motorway" OFF by default? In most
+		countries it is absolutely forbidden to use them with a bicycle. So we should
+		make sure that the user can only turn this option on ACTIVELY. Maybe even a pop up warning? -->
```
xmd5a2: Are you sure it is ON by default now? Option was changed from avoid_motorways to allow_motorways because of that.

routing/routing.xml
Posteinwurf: I did not know that reason. That seems to make sense.

routing/routing.xml
Posteinwurf: Sorry, my comment was misleading. I did not want to say it is on by default now. Just wanted to make sure that it is off in any case.  If you already changed it and it works, then everything fine.

routing/routing.xml
```diff
@@ -549,7 +561,15 @@
 			<select value="-1" t="mtb:scale" v="5"/>
 			<select value="-1" t="mtb:scale" v="6"/>
 
-			<select value="-1" t="vehicle" v="no"/>
+			<select value="1" t="vehicle" v="no"/>
+			<!-- The tag vehicle=no is normally used when mappers
+			want to exclude motor vehicles. I have never seen a case
+			when such a road was actually forbidden for bicycles as
+			well. (Mappers often do not bother to add the bicycle=yes).
+			Example:www.openstreetmap.org/way/159679495  This is
```
xmd5a2: This is wrong tagging. According to wiki vehicle=no forbids way for all vehicles (motorized and non-motorized). Also, access=no forbids access to every person (vehicle, pedestrian etc). So I disagree. Proper tagging on this way should be motor_vehicle=no if I understand mapper's intentions correctly.

routing/routing.xml
```diff
@@ -614,12 +630,22 @@
...
-			</if>
+			<!-- <select value="22" t="bicycle" v="designated"/>-->
+			<!-- Line above: Does the "bicycle designated" work here?
+			So far, we have assigned a kmh value to tracks, paths etc - 
+			can a kmh value also be assigned to a tag like "bicycle designated"?
+			If so: Would all tracks, paths etc tagged with "bicycle designated"
+			be multiplied later on (so that, if they are tagged with "paved" or
+			"asphalt", the 23 kmh would be multiplied with the "paved" factor)?
+			Could someone please help me understand this?
+			If it works as described, it adds some great possibilities.-->
```
xmd5a2: No, It doesn't work that way. This line was added because almost always ways tagged with bicycle=designated have good surface. I tried to increase their priority but I agree this is not the best solution. But lets keep this until better solution is found.

routing/routing.xml
```diff
@@ -614,12 +630,22 @@
...
+
+			<!-- <select value="4" t="highway" v="track"/> <select value="4" t="highway" v="footway"/> <select value="3" t="highway" v="path"/>-->
+			<!-- Line above: This line contained downgrades for the
+			"avoid unpaved" mode and has been moved further down.
+			Osmand first has to check if a path or track is paved,
+			asphalt, etc. Only if no tag like "asphalt" has been
+			found, then paths / tracks etc should be downgraded.-->
```
xmd5a2: If we have a perfectly filled up OSM database where every highways have surface, smoothness etc tags then yes. This section was introduced because most tracks doesn't have even surface tag and most of them are unpaved. It applies to "path" too. Footway should be mistake. In my tests without this section avoid_unpaved option doesn't do a good job.

routing/routing.xml
Posteinwurf: Actually, the problem of left turns is a very good point. 

Is there any possibility to add a penalty _precisely_ for that (only left turns, only on main roads)? Maybe we can make something up? That would be great. 

At the moment it is a dilemma: If the routing engine is programmed to stay away from main roads, it will sometimes do so by exiting a main road to the left. The maps by themselves normally do not give any hint: Often the left turn is just done be entering into a residential road.  https://www.openstreetmap.org/way/66514469 Since there is no "crossing" marked, or anything that could be applied with a penalty, it would be great to just be able to put a penalty on ANY left turn on main roads.

routing/routing.xml
Posteinwurf: I did want to suggest to delete it - just suggested to move it further down. 

Once we have checked that there is no asphalt etc, the paths, tracks, footways will be heavily downgraded in "unpaved" mode (please look further down - I think it makes sense there).

xmd5a2: Yes, I've done some local tests with your routing.xml and this change seems not to break anything.

routing/routing.xml
Posteinwurf: Yes, you are right concerning the definition. The problem is, that wrong tagging is the norm with vehicle=no. Maybe this is because in everyday language, the term "vehicle" is not used for bicycles at all. If you do a picture search for "vehicle" it only produces cars, lorries, tanks etc

When you go with a bike, you can always dismount. So maybe we can allow vehicle=no here and give it a factor of 0.4 later?

routing/routing.xml
Posteinwurf: Thats a good idea. I did not think of that. So when we put it here, it will just get a fixed value, without any other multiplication? 

In that case, we should even set it even higher, maybe 35 kmh. Because the multiplications can lead to very high numbers as well (for example a track with bicycle=official or a residential road with a cycle track).

routing/routing.xml
Posteinwurf: I did not move it much further down. The only difference should be, that any track etc that is clearly marked "asphalt" will not be downgraded in "avoid unpaved" mode.

routing/routing.xml
xmd5a2: Yes, no multiplication.

routing/routing.xml
xmd5a2: Yes, it can be a good compromise.

Posteinwurf: I was wrong: It is the better idea to let "vehicle=no" as it was. It is only important to check "bicycle=yes" (etc) first, and put "vehicle=no" behind that. I have put it in like this in my updated pull request (that I put online today). 

The compromise (downgrading it later) does not work for the following reason: To make the downgrade work, we have to put it before "surface=asphalt". But in that case, the "bicycle=yes" comes behind the downgrade for "vehicle=no". So the very common case (where we have vehicle=no combined with bicycle=yes) would produce a wrong result: It would get a strong downgrade, even though bicycle=yes is marked. www.openstreetmap.org/way/288180634

xmd5a2:
Add footway=* to routing https://github.com/osmandapp/OsmAnd-resources/commit/baa88d6ab0a1e1d750ef08c6a67e2b90179ddead

PR #435, opened 2017-06-01
Revised pull request

Posteinwurf:
Dear all,

based on the discussion so far, I have created a new pull request. I have left out all the suggestions that do not seem necessary any more. 

But still there are same changes that would be helpful. Could you include these into the next Osmand update, please?

Would it be possible to also make Osmand able to "see" footway=sidewalk and highway=elevator while routing? The first one would even be an urgent update, I think. 

Also, the triple counting of traffic lights would be important to solve in my opinion. But this is obviously more complicated. 

Thank you very much

Malte

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -559,6 +559,7 @@
 			<select value="-1" t="sac_scale" v="difficult_alpine_hiking"/>
 			<select value="-1" t="mtb:scale" v="5"/>
 			<select value="-1" t="mtb:scale" v="6"/>
+			<!-- vehicle=no is now in line 577 -->
 			<select value="1" t="vehicle" v="agricultural"/>
 			<select value="1" t="vehicle" v="forestry"/>
 			<select value="1"  t="vehicle" v="yes"/>
@@ -566,7 +567,6 @@
 			<select value="1"  t="vehicle" v="designated"/>
 			<select value="1"  t="vehicle" v="destination"/>
 			
-			<!-- vehicle=no is now in line 577 -->
 			<select value="-1" t="access" v="no"/>
 			<select value="-1" t="access" v="private"/>
 			<select value="1" t="access" v="agricultural"/>
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -484,7 +484,7 @@
 	</routingProfile>
 
 	<routingProfile name="bicycle" baseProfile="bicycle" restrictionsAware="true" minDefaultSpeed="2" maxDefaultSpeed="37"
-			leftTurn="0" rightTurn="0" followSpeedLimitations="false" onewayAware="true" heuristicCoefficient="1.4">
+			leftTurn="4" rightTurn="2" followSpeedLimitations="false" onewayAware="true" heuristicCoefficient="1.4">
 		<!-- <attribute name="relaxNodesIfStartDistSmallCoeff" value="2.5"/>  -->
 		<!-- New ROUTING API -->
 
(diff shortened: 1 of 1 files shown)


routing/routing.xml
```diff
@@ -614,12 +624,22 @@
 			<select value="1.3" t="highway" v="steps"/>
 			<select value="3" t="route"   v="ferry"/>
 			<select value="5" t="bicycle" v="dismount"/>
-			<select value="22" t="bicycle" v="designated"/>
-			<if param="avoid_unpaved">
-				<select value="4" t="highway" v="track"/>
-				<select value="4" t="highway" v="footway"/>
-				<select value="3" t="highway" v="path"/>
-			</if>
+			<!-- Line above: I now think it would be a better solution
```
xmd5a2: What about a case where highway=primary etc. + (highway=footway+bicycle=designated) are parallel? In many countries it is not allowed to drive on main highway if it is possible to use cycleway.

Posteinwurf: Could you explain further (maybe even post a link to a map) so that I can think about that case?

Posteinwurf:
Just to let you know: The long distance biker that I contacted this week is already a big fan of Osmand bicycle routing! He has just started using it and he told me: "The IT engineer who sang the praises of OsmAnd to me in Cremona was right." That sounds good, doesn't it?

I think Osmand is already the best bicycle routing engine now. Especially with the elevation data inclusion, and also the driving styles that deliver quite different results (suiting everyone's needs). All the other programs have their flaws: Much too complicated (Brouter), far too many main roads (Google), only limited coverage of areas (Bike Citizens), etc. 

I hope that we can soon eliminate the last little flaws. Afterwards, I could make some PR work for Osmand as a bike routing device (my main job is working as a journalist, so I would have some ideas how to promote Osmand). I think Osmand will get even more popular with cyclists. 

The question that you brought up about left turns on main roads still seems very important to me. It would be great to find a solution here. Ideally a solution that puts a penalty only on those left turns done on main roads. You think that would be doable? Or another thought: Maybe just any crossing of a main road could get a penalty? (That would include crossings where you approach the main road straight on, without a turn). 

Unfortunately, the OSM maps very often do not help here: Many crossings are not marked as a crossing at all (so no possibility to put a penalty on a tag like "crossing"). Take this, for example: www.openstreetmap.org/way/334070741 It just says "footway" (so there is no "crossing=island" or a similar tag that could get a penalty). But maybe it is doable within Osmand to mark the crossings as a crossing and put a slight penalty on them?

Best wishes

Malte

routing/routing.xml
xmd5a2: If we remove bicycle=designated here, will OsmAnd route via main highway it this case (with "Fast bike" driving style)?

xmd5a2:
Fixed highway=elevator https://github.com/osmandapp/OsmAnd-resources/commit/e933e61755b32ed762e9f24643d999907ce9114f

routing/routing.xml
Posteinwurf: Yes, you are absolutely right. Good point. The problem exists in driving style "speed" (not in the other two style). There is a smooth solutions for this: 

We just move the fixed kmh value for bicycle=designated a few lines further down, into the exceptions for the driving style "speed". I have already put this into the pull request. I have also deleted the remark about elevators (thanks for changing that in the map rendering).

Posteinwurf:
Thank you for fixing the elevator thing! Could you do the same for  footway=sidewalk ? That is even more important. I hope it can be done also without too much hassle.

Posteinwurf:
I have thought again about the "left turns on main roads" problem. I think, the best solution might be to have a penalty for ANY crossing of a main road. Because now, a crossing a main road gets a) no penalty at all (if OSM just shows a street or footway joining into the main road) b) a slight penalty OR c) a heavy penalty (some crossings over two lane main roads are marked with two times highway=crossing PLUS one time crossing=island), etc.  

I think, it would make sense to  apply a penalty of 15 or even 20 seconds for any crossing of a main road. That would also include left turns on main roads (hopefully). Is there any way of implementing this?

If this is solved, it would also partially solve the problem of triple counting of traffic lights (we could then just set the values for traffic lights much lower, because they already get a penalty if they lead over a main road).

PR #442, opened 2017-07-11
Some additional improvements

Posteinwurf:
I did a revision of the routing.xml
Please check it and include it in the July update, if still possible.

The proposed changes are based on much testing and research (for example much research on the quality of bicycle routes)

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -658,11 +658,17 @@
 			<!-- Line above: This tag is used for cycleways along main
 			roads. These need to be downgraded before the 31 kmh
 			for bicycle designated are set. -->
-			<select value="31" t="bicycle" v="designated"/>
-			<!-- Lines above: Sorry - I made a mistake in the former version. It makes sense to set
-			a fixed value here for bicycle=designated.  Otherwise there would be one major error:
-			The combination of designated AND asphalt (frequently found ) would get a much
-			lower value than just designated alone. -->
+			
+			<ifnot param="avoid_unpaved">
+				<select value="31" t="bicycle" v="designated"/>
+			</ifnot>
+			
+			<!-- Lines above: This fixed value for bicycle=designated will not be applied in 
+			"avoid unpaved" mode. Reason: The kmh value that we set here is fixed, it can not
+			be downgraded later. Generally, it makes sense to set a fixed value here for
+			bicycle=designated.  Otherwise there would be one major error: The combination of
+			designated AND asphalt (frequently found ) would get a much lower value than just
+			designated alone. -->
 
 			<if param="driving_style_safety">
 				<select value="7" t="highway" v="primary"/>
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -483,7 +483,7 @@
 		</point>
 	</routingProfile>
 
-	<routingProfile name="bicycle" baseProfile="bicycle" restrictionsAware="true" minDefaultSpeed="2" maxDefaultSpeed="37"
+	<routingProfile name="bicycle" baseProfile="bicycle" restrictionsAware="true" minDefaultSpeed="2" maxDefaultSpeed="33"
 			leftTurn="4" rightTurn="2" followSpeedLimitations="false" onewayAware="true" heuristicCoefficient="1.4">
 		<!-- <attribute name="relaxNodesIfStartDistSmallCoeff" value="2.5"/>  -->
 		<!-- New ROUTING API -->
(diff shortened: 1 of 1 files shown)


maraf24:
So highway=cycleway + vehicle=no would be routeable? Why?

Posteinwurf:
several reasons:

a) vehicle=no is tagged wrongly very often. it indicates ways that are only forbidden for cars. but mappers very often forget to add the additional tag bicycle=yes. I habe checked several cases and found this error frequently. Once it was even an official bicycle route that was tagged with vehicle=no.

b) even the authorities sometimes forget to add the sign that indicates the exception for cycles. there are ways where on one side the exception for cycles is indicated, but on the other side it is not.

c) if you really hit on a road that is not allowed for cycles, you can still push your bike. if it is not allowed to push, it will be access=no which is excluded in the routing.xml

and lastly, with my suggestion, we still exclude many of the vehicle=no roads. they are only allowed if there is some hint that they are only forbidden for motor vehicles.

i think it is tagged wrongly quite often, because in common language, people would not think of a bike when they are talking of a vehicle. if you do a picture search for vehicle, you get cars, lorries, tanks but no bicycles.

-- 
Diese Nachricht wurde von meinem Android Mobiltelefon mit WEB.DE Mail gesendet.

maraf24:
cycleway with vehicle=no should not appear often, but it it may still correct tagging. There are many possibilities:
1. it could be footway, not cycleway, ie. the highway tag is wrong, not access key.
2. it could be cycleway, which for some reason (nearby construction) is temporary footway. The mapper could decide to leave the highway=cycleway unchanged for future reference. One should look for foot tag to confirm it.

Nonetheless, we should not try to guess and correct tagging errors in routing.xml.  It's a task for QA tool and ground survey.

Posteinwurf:
Please do allow me some more general remarks about bicycle routing:

I might be the person in the team that tested the bicycle routing most intensively. I have spent entire weeks finding out which values lead to a great routing result. This included evaluations of the OSM database via Overpass and many field tests in different countries (I have to travel a lot for my job, and I normally take a bike to my destination). I have tested in Vienna, Prague, Hamburg, Berlin, Paris, Cologne, Southern France, Poland, Lithuania, and Northern and Southern Germany.

After all this research, I think that there is one difference between bicycle routing and car routing. In car routing, normally you know for sure if a way in the OSM database is suitable. In bicycle routing, sometimes you do not know, because it is mapped inaccurately (or even wrongly). This is because there are so many ways that are potentially suitable for bicycle routing (all the paths for example), and many of them are not accurately mapped (many of those small ways in parks, fields, forests, etc.)

Of course, even in bicycle routing, most of the time you will be on a way with a clear indication that it is good for bicycles (most residential roads for example). But _in a few cases_ you need to use heuristics. By doing so, you get a much better routing result: Think of all those paths that have no additional tag at all. If such a path provides a real shortcut, it is a good idea to use it. In more than 99 per cent of cases this saves time. And, even more important: It will probably be safer, because it helps you to avoid dangerous main roads. Only in some rare cases, you might then end up on a path where you have to push your bike (bad surface, or only allowed for pedestrians, etc). If we would exclude all paths that have no explicit bicycle=yes or bicycle=designated, we would loose some really great shortcuts.

Of course, vehicle=no is a more extreme case, because formally, someone has set a tag here that is meant to exclude bicycles. But I have done some resarch: Vehicle=no is used wrongly very often: It indicates exactly those ways that are only forbidden for cars (and thus are most suitable for bicycles). Some examples (all these are marked vehicle=no combined with cycleway).

grafik

https://www.openstreetmap.org/way/194975969
https://www.openstreetmap.org/way/161734787
https://www.openstreetmap.org/way/23986853
https://www.openstreetmap.org/way/132043166
https://www.openstreetmap.org/way/35206805
https://www.openstreetmap.org/way/25444640
https://www.openstreetmap.org/way/24880691

All these examples are important shortcuts - if they are included, the routing gets better. So my suggestion is: If vehicle=no is combined with a hint that indicates bicycles are allowed, we should in fact include this way. One of theses hints is a combination of "cycleway" and vehicle=no. 

To sum it up: I know it seems a bit weird to use heuristics in routing. But in some rare cases this makes sense, because we get a much better routing result (great shortcuts and ways with less car traffic). And we should not be too anxious about it: The real danger is not putting a cyclist on a way where he might have to dismount and push for some time. The real danger is setting him on main roads with heavy traffic and no bike path (this might be totally legal, but at the same time deadly dangerous).

To put it in perspective: During all the weeks when I have tested routing, I just had _three_ cases where I had to dismount and push. All of them due to bad surface. So all the ways that the routing engine suggested during this time were legal for bicycles (and sometimes I even used quite extreme heuristics during my tests).

Best wishes

Malte

Posteinwurf:
Thank you for merging!


I am currently working on some improved settings for elevation. So if the team is preparing to put out an update – could you please hold back for another hour or so?

Have been testing for some hours now, will be ready soon!

PR #443, opened 2017-07-13
Add support for showing GPX tracks in pink

njohnston:
Pink is one of the least used colours on the map, making it a good choice for clearly showing GPX tracks. Pink was previously available in OsmAnd 2.5 but removed in 2.6.


njohnston:
@xmd5a2: thanks, could you please merge the associated OsmAnd pull request (https://github.com/osmandapp/Osmand/pull/4107) too?

PR #444, opened 2017-07-17
refined values for elevation in bicycle mode

Posteinwurf:
Dear all,

I have improved the values for the elevation mode and added them to the routing.xml. I did many hours of virtual testing in Vienna, Berlin and Hamburg. Before, the elevation mode sometimes would suggest a bad route, just to save 5 or 10 height meters. Now, it is more balanced (especially if we consider the "noise" that is in the elevation data). Also, now you get a clearer diffence between the three elevation modes (most leveled really suggesting the route with least elevation, etc).

The revision is based on one main thought: If you have to climb 1000 meters, it is very bad at 13 per cent incline, but still quite bad at 3 per cent incline. You still have to climb these 1000 meters (so also with 3 per cent incline, it is a good idea to take a longer detour to avoid them). According to my tests, the emphasis should first be on avoiding the height meters, and after that on reducing the incline. Even with the new values, a high incline can double or triple the penalty (in some cases even more than that). So first, height meters are avoided. But if they cannot be avoided, the routing engine will still heavily prefer a route where these height meters are better 'distributed" over the whole route (so, same height to climb, but less incline). 

I will do some further testing, but please do already include this into the routing.xml, because according to my thourough testing, it is a big step forward.

If someone else wants to do some testing: A good starting point for a "calibration" are these two routes

a) From Marktgasse 21 (Vienna) to ORF-Zentrum (Kueniglberg, Venna). In safety mode on the last uphill section, the routing should take Wattmanngasse (residential) not the Maxingstrasse (tertiary). The routing over Maxingstrasse is supposed to have 7 height meters less (that could also just be noise in the data). Only to save 7 metres, it does not make sense to go 500 metres uphill on a tertiary road without a bike path (this only makes sense in "speed" mode).  If the routing takes Maxingstrasse in "safety" mode, this is a sign that the elevation gets too much weight compared to the other OSM data.

b) From the public swimming pool in Huetteldorf (Vienna) to Langenlebarn Volksschule. In "most leveled", the routing should make a detour that has an additional 13,5 kilometres, but at the same time has 164 height meters LESS. (This does not apply in "fast" mode). This route is good for checking the routing engine's preferences, because all three main alternatives have their shortcomings: The shortest route has a long part over secondary and tertiary road without any bike path. A slightly longer route is much safer, but has more elevation and a 4km section of track with unknown surface. And then there is the third alternative that is very safe and very flat, but much longer. 

Greetings from Berlin,

Malte

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -727,8 +727,17 @@
 			<if param="height_obstacles">
 
 				<gt value1="-100" value2=":incline">
-                	                <select value="-1"/>
-                        	</gt>
+					<select value="1000" t="surface" v="cobblestone"/>
+                                        <select value="800" t="surface" v="paved"/>
+                                        <select value="800" t="surface" v="asphalt"/>
+                                        <select value="800" t="surface" v="concrete"/>
+                                        <select value="1000" t="osmand_highway_integrity_brouting_low" v="yes"/>
+                                        <select value="1000" t="highway" v="path"/>
+                                        <select value="1000" t="highway" v="track"/>
+                                        <select value="1000" t="highway" v="footway"/>
+                                        <select value="800"/>
+				</gt>
+
 				<!-- Above: Downhill between 100 and 60 is so steep that it should probably
 				be excluded in any case, even in hilly mode and for paved roads.-->
 
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -1058,7 +1058,10 @@
 			leads to routing errors. The above is only a compromise to minimize
 			those errors.-->
 			
-			<select value="9" t="crossing" v="traffic_signals"/>
+			<if param="driving_style_speed">
+				<select value="8" t="crossing" v="traffic_signals"/>
+			</if>
+			<select value="11" t="crossing" v="traffic_signals"/>
 			<select value="1" t="crossing" v="uncontrolled"/>
 			<select value="3" t="crossing" v="unmarked"/>
 			<select value="1" t="crossing" v="island"/>
@@ -1071,6 +1074,9 @@
 			to apply any penalty. So it would make no sense to just apply a high
 			penalty on some cases. -->
 			<select value="5" t="crossing" v="island;traffic_signals"/>
+			<if param="driving_style_speed">
+				<select value="11" t="highway" v="traffic_signals"/>
+			</if>
 			<select value="19" t="highway" v="traffic_signals"/>
 			<select value="55"  t="highway" v="steps"/>
 			<select value="15"  t="bicycle" v="dismount"/>
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -725,30 +725,41 @@
 		<way attribute="obstacle_srtm_alt_speed">
 			<if param="height_obstacles">
 				<gt value1="-100" value2=":incline">
+					<select value="-1" t="surface" v="cobblestone"/>
 					<select value="-1" t="osmand_highway_integrity_brouting_low" v="yes"/>
 					<select value="-1" t="highway" v="path"/>
 					<select value="-1" t="highway" v="track"/>
+					<select value="-1" t="highway" v="footway"/>
 				</gt>
 				<gt value1="-25" value2=":incline">
+					<select value="170" t="surface" v="cobblestone"/>
+					<select value="1" t="surface" v="paved"/>
+					<select value="1" t="surface" v="asphalt"/>
+					<select value="1" t="surface" v="concrete"/>
 					<select value="180" t="osmand_highway_integrity_brouting_low" v="yes"/>
 					<select value="180" t="highway" v="path"/>
 					<select value="170" t="highway" v="track"/>
+					<select value="180" t="highway" v="footway"/>
 				</gt>
 				<gt value1="-17" value2=":incline">
+					<select value="140" t="surface" v="cobblestone"/>
+					<select value="1" t="surface" v="paved"/>
+					<select value="1" t="surface" v="asphalt"/>
+					<select value="1" t="surface" v="concrete"/>
 					<select value="150" t="osmand_highway_integrity_brouting_low" v="yes"/>
 					<select value="150" t="highway" v="path"/>
 					<select value="140" t="highway" v="track"/>
+					<select value="150" t="highway" v="footway"/>
 				</gt>
 				<gt value1="-9" value2=":incline">
+					<select value="1" t="surface" v="paved"/>
+					<select value="1" t="surface" v="asphalt"/>
+					<select value="1" t="surface" v="concrete"/>
 					<select value="30" t="osmand_highway_integrity_brouting_low" v="yes"/>
 					<select value="30" t="highway" v="path"/>
 					<select value="25" t="highway" v="track"/>
 				</gt>
-				<gt value1="-3" value2=":incline">
-					<select value="2" t="osmand_highway_integrity_brouting_low" v="yes"/>
-					<select value="2" t="highway" v="path"/>
-					<select value="2" t="highway" v="track"/>
-				</gt>
+
 				<if param="relief_smoothness_factor_more_plains">
 					<gt value1="4" value2=":incline">
 						<select value="5"/>
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -723,90 +723,157 @@
 		</way>
 
 		<way attribute="obstacle_srtm_alt_speed">
+
 			<if param="height_obstacles">
+
 				<gt value1="-100" value2=":incline">
-					<select value="-1" t="surface" v="cobblestone"/>
-					<select value="-1" t="osmand_highway_integrity_brouting_low" v="yes"/>
-					<select value="-1" t="highway" v="path"/>
-					<select value="-1" t="highway" v="track"/>
-					<select value="-1" t="highway" v="footway"/>
+                	                <select value="-1"/>
+                        	</gt>
+				<!-- Above: Downhill between 100 and 60 is so steep that it should probably
+				be excluded in any case, even in hilly mode and for paved roads.-->
+
+				<gt value1="-60" value2=":incline">
+					<select value="400" t="surface" v="cobblestone"/>
+                                        <select value="300" t="surface" v="paved"/>
+                                        <select value="300" t="surface" v="asphalt"/>
+                                        <select value="300" t="surface" v="concrete"/>
+                                        <select value="400" t="osmand_highway_integrity_brouting_low" v="yes"/>
+                                        <select value="400" t="highway" v="path"/>
+                                        <select value="400" t="highway" v="track"/>
+                                        <select value="400" t="highway" v="footway"/>
+                                        <select value="330"/>
 				</gt>
-				<gt value1="-25" value2=":incline">
-					<select value="170" t="surface" v="cobblestone"/>
-					<select value="1" t="surface" v="paved"/>
-					<select value="1" t="surface" v="asphalt"/>
-					<select value="1" t="surface" v="concrete"/>
-					<select value="180" t="osmand_highway_integrity_brouting_low" v="yes"/>
-					<select value="180" t="highway" v="path"/>
-					<select value="170" t="highway" v="track"/>
-					<select value="180" t="highway" v="footway"/>
-				</gt>
-				<gt value1="-17" value2=":incline">
-					<select value="140" t="surface" v="cobblestone"/>
-					<select value="1" t="surface" v="paved"/>
-					<select value="1" t="surface" v="asphalt"/>
-					<select value="1" t="surface" v="concrete"/>
-					<select value="150" t="osmand_highway_integrity_brouting_low" v="yes"/>
-					<select value="150" t="highway" v="path"/>
-					<select value="140" t="highway" v="track"/>
-					<select value="150" t="highway" v="footway"/>
-				</gt>
-				<gt value1="-9" value2=":incline">
-					<select value="1" t="surface" v="paved"/>
-					<select value="1" t="surface" v="asphalt"/>
-					<select value="1" t="surface" v="concrete"/>
-					<select value="30" t="osmand_highway_integrity_brouting_low" v="yes"/>
-					<select value="30" t="highway" v="path"/>
-					<select value="25" t="highway" v="track"/>
+				<!-- Lines above: This is still so steep, that a high penalty makes sense in any mode, and
+                                even on paved surface. Maybe the values need some refinement.-->
+
+				<gt value1="-40" value2=":incline">
+                                        <select value="200" t="surface" v="cobblestone"/>
+					<select value="100" t="surface" v="paved"/>
+                                        <select value="100" t="surface" v="asphalt"/>
+                                        <select value="100" t="surface" v="concrete"/>
+                                        <select value="200" t="osmand_highway_integrity_brouting_low" v="yes"/>
+                                        <select value="200" t="highway" v="path"/>
+                                        <select value="200" t="highway" v="track"/>
+                                        <select value="200" t="highway" v="footway"/>
+                                        <select value="150"/>
 				</gt>
 
+				<gt value1="-25" value2=":incline">
+					<select value="140" t="surface" v="cobblestone"/>
+					<select value="40" t="surface" v="paved"/>
+					<select value="40" t="surface" v="asphalt"/>
+					<select value="40" t="surface" v="concrete"/>
+					<select value="150" t="osmand_highway_integrity_brouting_low" v="yes"/>
+					<select value="150" t="highway" v="path"/>
+					<select value="150" t="highway" v="track"/>
+					<select value="150" t="highway" v="footway"/>
+                                        <select value="60"/>
+				</gt>
+
+				<gt value1="-20" value2=":incline">
+					<select value="40" t="surface" v="cobblestone"/>
+					<select value="1" t="surface" v="paved"/>
+					<select value="1" t="surface" v="asphalt"/>
+					<select value="1" t="surface" v="concrete"/>
+					<select value="80" t="osmand_highway_integrity_brouting_low" v="yes"/>
+					<select value="80" t="highway" v="path"/>
+					<select value="80" t="highway" v="track"/>
+                                        <select value="1"/>
+				</gt>
+
+				<gt value1="-9" value2=":incline">
+					<select value="0.1" t="surface" v="paved"/>
+					<select value="0.1" t="surface" v="asphalt"/>
+					<select value="0.1" t="surface" v="concrete"/>
+					<select value="30" t="osmand_highway_integrity_brouting_low" v="yes"/>
+					<select value="30" t="highway" v="path"/>
+					<select value="30" t="highway" v="track"/>
+                                        <select value="0.1"/>
+				</gt>
+
+
+			
 				<if param="relief_smoothness_factor_more_plains">
 					<gt value1="4" value2=":incline">
+						<select value="3"/>
+					</gt>
+					<gt value1="8" value2=":incline">
 						<select value="5"/>
 					</gt>
-					<gt value1="7" value2=":incline">
-						<select value="7"/>
-					</gt>
 					<gt value1="15" value2=":incline">
-						<select value="9"/>
+						<select value="35"/>
 					</gt>
 					<gt value1="25" value2=":incline">
-						<select value="12"/>
+						<select value="120"/>
 					</gt>
-					<select value="4"/>
-				</if>
-				<if param="relief_smoothness_factor_hills">
-					<gt value1="4" value2=":incline">
-						<select value="2.5"/>
+					<gt value1="30" value2=":incline">
+						<select value="240"/>
 					</gt>
-					<gt value1="7" value2=":incline">
-						<select value="3.5"/>
-					</gt>
-					<gt value1="15" value2=":incline">
-						<select value="4.5"/>
-					</gt>
-					<gt value1="25" value2=":incline">
-						<select value="6"/>
+					<gt value1="35" value2=":incline">
+						<select value="480"/>
 					</gt>
 					<select value="1"/>
 				</if>
 
-				<gt value1="4" value2=":incline">
-					<select value="3.5"/>
-				</gt>
-				<gt value1="7" value2=":incline">
-					<select value="4.5"/>
-				</gt>
-				<gt value1="15" value2=":incline">
-					<select value="6"/>
-				</gt>
-				<gt value1="25" value2=":incline">
-					<select value="8"/>
-				</gt>
-				<select value="1.5"/>
+				<if param="relief_smoothness_factor_hills">
+
+					<gt value1="1" value2=":incline">
+						<select value="0.1"/>
+					</gt>
+					<gt value1="8" value2=":incline">
+						<select value="0.3"/>
+					</gt>
+					<gt value1="15" value2=":incline">
+						<select value="4"/>
+					</gt>
+					<gt value1="25" value2=":incline">
+						<select value="70"/>
+					</gt>
+					<!-- Incline above 25 per cent should get a high penalty, even
+					in hills mode: The steepest incline at the Tour de France in
+					2015 was 27 per cent. That was the peak value, while the
+					average value of that incline was 10 per cent. Most 
+					famous Tour de France inclines only have average values
+					of 10 to 15 per cent.
+					The higher inclines need much higher penalties, because 
+					they could appear only once in a longer route. So if you really want
+					to make a difference for this one incline, it needs a high (single)
+					penalty. An incline of 10 per cent can appear many times, and will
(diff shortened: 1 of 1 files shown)


xmd5a2:
At first sight I was impressed by calculation speed difference but soon found absolutely inadequate route.
Location: http://www.openstreetmap.org/?lat=44.5279531785173&lon=34.114031849831&mlat=44.5279531785173&mlon=34.114031849831&zoom=13
Map: http://builder.osmand.net/indexes/Russia_crimea_asia_2.obf.zip
Your routing produces this route (look at left-top graph - it is osmand generated slope/distance graph): https://www.dropbox.com/s/s9r1xa6nsk82d73/Screenshot%20from%202017-07-17%2020-15-40.png?dl=0
Old routing: https://www.dropbox.com/s/2jfd9keuie2mxi4/Screenshot%20from%202017-07-17%2020-16-09.png?dl=0
Route calculated with your routing is 2 times longer but at the beginning it leads downhill ~800m and then uphill ~800m. Original route height variations is much lower:
screenshot from 2017-07-17 20-26-16
Please note that osmand map creator generates route.gpx and route_slope.gpx after each route generation. In route_slope.gpx height values replaced by slope values. This way slope can be visualised by almost any track analyser.

Posteinwurf:
Thank you for testing it! I will check and try to revise the values.

I have just compared the calculation times myself, and could not find a difference. It was done on a virtual device Galaxy S6 (so times will be longer on most devices - but the differences between the versions should be the same as in my test:

Current Osmand version (via Google Play Store) 44 to 48 seconds (depending on mode)
Current routing.xml (without this pull request) 21 to 25 seconds
Revised routing.xml (with this pull request) 21 to 25 seconds

Posteinwurf:
Have just downloaded the Krim area. Will test. Great area for testing.

xmd5a2:
Looks like I actually forgot to add height_obstacles option while testing your routing.xml. Now routes are very similar.

Posteinwurf:
ok, I got the same result. In any mode that I tested, the routing results were the same as yours. 

grafik

Posteinwurf:
I have to recheck the last two commits. I have a better idea. So leave them aside at the moment. Will revise tomorrow (Tuesday).

routing/routing.xml
```diff
@@ -725,76 +725,85 @@
...
 				<gt value1="-100" value2=":incline">
+					<select value="-1" t="surface" v="cobblestone"/>
 					<select value="-1" t="osmand_highway_integrity_brouting_low" v="yes"/>
 					<select value="-1" t="highway" v="path"/>
 					<select value="-1" t="highway" v="track"/>
+					<select value="-1" t="highway" v="footway"/>
 				</gt>
 				<gt value1="-25" value2=":incline">
+					<select value="170" t="surface" v="cobblestone"/>
+					<select value="1" t="surface" v="paved"/>
```
xmd5a2: Why did you added these lines with value="1"?

xmd5a2:
I've tested this pull request more and I think that we need to increase penalties for high slopes. Look at this route
http://brouter.de/brouter-web/#zoom=12&lat=44.447&lon=33.8454&layer=OpenStreetMap&lonlats=33.885269,44.477891|33.975735,44.39675&nogos=&profile=fastbike&alternativeidx=0&format=geojson
screenshot from 2017-07-18 20-23-10
Slope profile:
screenshot from 2017-07-18 20-18-40
Parameters: bicycle,height_obstacles,allow_motorway,avoid_ferries,driving_style_speed,relief_smoothness_factor_plains
This is not what user wants when selecting relief_smoothness_factor_plains

Posteinwurf:
Thank you for testing it! I appreciate that. I have not tested the height mode as thoroughly as I normally do. This is because there are some bugs in the existing elevation mode that should be fixed. I hope to include these bugfixes it in the next update (do you know when it is due?)

 

I will try to look into your example and refine the values.

Posteinwurf:
Tracks and paths are sometimes marked with “asphalt”. I wanted to make sure that the penalty for bad downhill tracks is just applied to tracks / paths that are not paved.

routing/routing.xml
Posteinwurf: Yes, tracks and paths are sometimes marked with “asphalt”, "paved" or "concrete". I wanted to make sure that the penalty for bad downhill tracks is not applied to tracks / paths that are clearly marked as paved.

Posteinwurf:
Have looked into the example and cannot get clear values. Brouter shows a different elevation profile than my Osmand, and my Osmand shows different profile than your tests. Will do more testing tomorrow.

grafik

xmd5a2:
Swap start and finish but this is not main point. I know this place and original route is much easier than going uphill near Foros.

xmd5a2:
Also note that slope graph from OsmAnd itself is filtered as opposed to raw values from route_slope.gpx obtained via OsmAndMapCreator.

Posteinwurf:
Alright, now I see it! With swapped start and end point, Osmand shows 45% slope on one point...

xmd5a2:
In places of steep curves of a mountain road such as serpentine, these slope values are incorrect. This is due to the insufficient accuracy of the altitude data. This can't be avoided without some sort of complicated heuristics. But as a whole it is visible that it is far not a plain and the original route is more flat.

Posteinwurf:
Thanks again for checking. Yes we need to change the values. I am already experimenting with them. Just let me check if I get the logic right: 


gt value1="4" value2=":incline"
select value="6"

This means: If the incline is more than 4 per cent, multiply the height meters with five. 
So, if we have got a part of the route that has 100 meters elevation with 4 per cent slope, we multiply, and get a penalty of 600 seconds for that part. Right?

Also very important: Is the order of the commands important? So if we have already had the two lines above, did we already define a value for every incline higher than 4 per cent? And does that mean that it will be useless to put the following behind: gt value1="7" value2=":incline" ?

Posteinwurf:
I have tracked the problem down a bit: It seems to work with the following values
value 16 for 25 per cent incline
value 30 for 25 per cent incline
(rest as before)

BUT: It only seemed to work in when I put these values in directly (for any elevation mode, hilly, leveled, most leveled). So maybe there is something wrong with the selection of the three elevation modes. Or I am getting something completely wrong...

Could you, Leonid or someone else also answer my questions above? Thank you!

Posteinwurf:
PS: Beautiful landscape...
grafik

xmd5a2:
> This means: If the incline is more than 4 per cent, multiply the height meters with 6.
So, if we have got a part of the route that has 100 meters elevation with 4 per cent slope, we multiply, and get a penalty of 600 seconds for that part. Right?
Right.

> Also very important: Is the order of the commands important? So if we have already had the two lines above, did we already define a value for every incline higher than 4 per cent? And does that mean that it will be useless to put the following behind: gt value1="7" value2=":incline" ?
Order matters. If gt value1="7" if before gt value1="4" then only gt value1="7" will be executed and gt value1="4" will never be executed.

xmd5a2:
Release 2.7 is planned in next 2 weeks

Posteinwurf:
thank you very much for pointing to that example! 
The values have to be much more spreaded than I thought. Now I have got a solution that seems to work in  the "more plains" mode. 

<gt value1="4" value2=":incline">
						<select value="4"/>
					</gt>
					<gt value1="8" value2=":incline">
						<select value="6"/>
					</gt>
					<gt value1="15" value2=":incline">
						<select value="35"/>
					</gt>
					<gt value1="25" value2=":incline">
						<select value="70"/>
					</gt>
					<gt value1="30" value2=":incline">
						<select value="130"/>
					</gt>

					<select value="3"/>
thank you very much for pointing to that example! 
The values have to be much more spreaded than I thought. Now I have got a solution that seems to work in  the "more plains" mode. This seems to work satifactorily in all our test cases so far: 

gt value1="4" value2=":incline" select value="4"
gt value1="8" value2=":incline"  select value="6"
gt value1="15" value2=":incline" select value="35"
gt value1="25" value2=":incline" select value="70"
gt value1="30" value2=":incline" select value="130"
select value="3"

Will do more testing tomorrow

Malte

Posteinwurf:
Just a general observation (also for anyone who will later try to improve the elevation calculation). 
One part of Leonids route is especially tricky, and shows the problems with serpentines: It is pretty clear here that the longer route via the secondary road (red colour) is much flatter. But the routing always prefers the shortcut via the path (according to the grafical display has an inlince of 40 per cent. Even if that were not true: it is clear that the path must be much steeper than the serpentines). 

BUT: No matter which penalty I put in for elevation (even quite extreme values) - the routing engine will ALWAYS take the path. The probable explanation: The elevation data for the serpentines are probably wrong in this case (obviously they have even more extreme then the shortcut via the path). So if we want to develevop a solution for this in the future, it might be a good test case.

grafik

xmd5a2:
I already mentioned this:
`In places of steep curves of a mountain road such as serpentine, these slope values are incorrect. This is due to the insufficient accuracy of the altitude data. This can't be avoided without some sort of complicated heuristics. `
Here is a QGIS screenshots of used elevation data and a profile. I think it is self explanatory. I see no way to calculate correct slopes with such a low resolution data for serpentine roads.
screenshot from 2017-02-26 19-21-09
screenshot from 2017-02-26 19-31-22

Posteinwurf:
Thank you for the explanatory data! That makes it clear.
I had already seen your earlier comment. My observation was intended as an example for your remark about serpentines.

P.S. I am almost ready to post new suggestions for values for incline / elevation. 

Do you think the flatter route (via 67k1) should be also taken in "hill" mode? I think that could make sense: When you have in mind that - according to the OSM data - the shorter route is considered to be even steeper than it actually is... So I think, if we have a road that is REALLY as steep as this, it should be avoided in all modes, even hilly, right?

xmd5a2:
Yes, I think you are right.

Posteinwurf:
Ok, will post in half an hour or so.  Thanks for your examples, I had not tested in such extreme territory (in the Alps, yes, but my test cases were easier to solve for Osmand, so I did not see the difficulties...). 

The steps above 25 per cent inline have to be much broader than I previously thought. More like the original values. 

I have put very low values in the lower ranges (up to 7 per cent incline). This is to avoid random results in cities with low inclines: There, you just want to change the route, if you can really save some height meters. If the penalty already gets high for low inclines, you get wrong routing caused by the "noise" in the data.

Posteinwurf:
P.P.S.: It is a pity that the data grid for mountains is so coarse. This means that all the (bad) tracks and paths that just go straigt up the mountain have a systematic advantage (in the data) over the (better) serpentines.

I have a very rough idea for the heuristics for serpentine calculation. But I do not know (at all) if this could be implemented or is just too complicated:

"If you, routing engine, detect a snake-like curve in a mountain region, then do not take any elevation data from in between: Just take the average elevation value from the beginning and from the end of the snake. Then compare it to the length of the that snake". 

This would not work for a single curve, but maybe for longer serpentines. These might be the main problem.

xmd5a2:
This is SRTM 30m, best available data with open license. But resolution of data in Alps will be improved. I've found such data and prepared it but don't want to put it to server until I finish ArcticDEM processing.

Posteinwurf:
Oh, great. Thank you for improving this!!!
I did not know about the different sources and that better data for the Alps will be available.

Posteinwurf:
PPPS: Could you just briefly check my understanding of the logics (in the following example)? I do want to get the systematics right. 

**Lets assume that I want to have the following factors:**
Factor 20 for any inline above 15 per cent
Factor 10 for any incline from 4-15 per cent
No penalty for any incline from 0 to 4 per cent
Exclusion of any incline (downhill) from -100 to -40 per cent
Factor 20 for any incline (downhill) from -40 to -10 per cent. 

**Now, to get the above, is the following order of commands right?**
gt value -40 select value -1 
gt value -10 select value 20 
gt value 4 select value 10
gt value 15 select value 20
select value 1

So will value 1 apply to any value between 0 and 4 then?
Or will it also apply to any value from -4 and 4?

Thank you!

Posteinwurf:
PPPS: Could you just briefly check my understanding of the logics (in the following example)? I do want to get the systematics right.

Lets assume that I want to have the following factors:
Factor 20 for any inline above 15 per cent
Factor 10 for any incline from 4-15 per cent
No penalty for any incline from 0 to 4 per cent
Total ban of any way with incline (downhill) from -100 to -40 per cent
Factor 20 for any incline (downhill) from -40 to -10 per cent.
Factor 2 for any incline (dowhill) from -10 to 0 per cent.

Now, to get the above, is the following order of commands right?
gt value -100 select value -1
gt value -40 select value 20
gt value -10 select value 2
gt value 4 select value 10
gt value 15 select value 20
select value 1

So will value 1 apply to any value between 0 and 4 then?
Or will it ALSO apply to any value from -10 and 0?
Or will it not apply at all, because we have already defined factor 2 for any value between -10 and 4?

Thank you!

I had to update this, therefore I deleted and reposted.

xmd5a2:
I think @vshcherb can answer better than me.

Posteinwurf:
thank you. Maybe I can also find out be testing with extreme values. But any help will be appreciated

Posteinwurf:
A comment on my commit from 5 minutes ago:

This is a suggestion for the elevation mode. The test cases seem to work well, but there is still one mistake, and I do not know why:

If I choose "fast" mode and "hilly", and calculate the route from "Langenlebarn Volksschule" (Austria) to Hallenbad Huetteldorf / Hütteldorf (Austria) - then the routing takes the long flat route along the Donau. 

But: In "hills", it should take the direct route via the mountains (15 km shorter, but only 50-80 height meters more with inclines between 5 and 15 per cent). If I reverse starting point and destination, it takes exactly that route - via the mountains (in hilly mode). The reversed variant is a bit more downhill, but still in "hilly" it should take the direct route in both directions. I have even introduced factors below 1 in hilly mode - nothing works. I do not understand why. The values in "hill" seem already to extreme to the low end. And even with much higher values, it would take the right route in the opposite direction.

vshcherb:
This is the code where it is used
```
double heightObstacle = ctx->config->router->defineHeightObstacle(road, !reverseWaySearch ? prevInd : segmentPoint, 
 					!reverseWaySearch ? segmentPoint : prevInd);
		if (heightObstacle < 0) {
			directionAllowed = false;
			continue;
		}
```
And here is the code for that function to be clear.
https://github.com/osmandapp/OsmAnd-core/blob/legacy_core/native/src/generalRouter.cpp#L321

routing/routing.xml
```diff
@@ -723,90 +723,157 @@
...
+
 			<if param="height_obstacles">
+
 				<gt value1="-100" value2=":incline">
-					<select value="-1" t="surface" v="cobblestone"/>
-					<select value="-1" t="osmand_highway_integrity_brouting_low" v="yes"/>
-					<select value="-1" t="highway" v="path"/>
-					<select value="-1" t="highway" v="track"/>
-					<select value="-1" t="highway" v="footway"/>
+                	                <select value="-1"/>
```
xmd5a2: We shouldn't completely forbid slopes < -100 because insufficient elevation data accuracy can produce even higher values at some hilly areas. These roads can be the only way to get to destination.

Posteinwurf:
ok, that makes sense!

Posteinwurf:
Dear Victor, 

thanks for the input! I still would like to have someone who explains it to me, because I do not understand that code. I could learn that language also, but I think the time might be better invested if I specialise in my area (which is testing the routing.xml values and finding the right combination for them).

So if someone could explain the example code values that I gave above, that would help a lot. 
And maybe someone could look into the example Langenlebarn to Huetteldorf (above)? I can give the exact starting and destination points if you want.

Maybe there is still one thing about the code that I do not understand.

vshcherb:
I think it is easier to understand it by example... Or may be I'm bad in explaining. Height obstacle is measured in seconds: vertical speed (defined in routing.xml) multiplied by the difference of height

Posteinwurf:
Thank you! 

But one part of my question is still unanswered: 

Will value 1 apply to any value between 0 and 4 then?
Or will it ALSO apply to any value from -10 and 0?
Or will it not apply at all, because we have already defined factor 2 for any value between -10 and 4?

**The base for this question is the following example**

**Lets assume that I want to have the following factors:**
Factor 20 for any inline above 15 per cent
Factor 10 for any incline from 4-15 per cent
No penalty for any incline from 0 to 4 per cent
Total ban of any way with incline (downhill) from -100 to -40 per cent
Factor 20 for any incline (downhill) from -40 to -10 per cent.
Factor 2 for any incline (dowhill) from -10 to 0 per cent.

**Now, to get the above, is the following order of commands right?**
gt value -100 select value -1
gt value -40 select value 20
gt value -10 select value 2
gt value 4 select value 10
gt value 15 select value 20
select value 1

Posteinwurf:
I am currently revising the elevation values, based on a 1000km tour that I just finished in France. 

But I need some input! There is something that I cannot explain at all. The folliwing two sets of values SHOULD produce the same routing. but they don't. They produce different results. Why? What am I getting wrong here?

FIRST SET OF VALUES:
       gt value1="4" value2=":incline"
             Select value="0.25"
          gt value1="17" value2=":incline"
             select value="0.25"
          gt value1="18" value2=":incline"
             select value="30"

SECOND SET IF VALUES:
        gt value1="4" value2=":incline"
             Select value="0.25"
          gt value1="18" value2=":incline"
             select value="30"

Posteinwurf:
ps I have left out some command details to make it more readable

Posteinwurf:
Could someone give me some input on the question that I have asked above? I will paste it here again: 

I am currently revising the elevation values, based on a 1000km tour that I just finished in France.

But I need some input! There is something that I cannot explain at all. The following two sets of values SHOULD produce the same routing. but they don't. They produce different results. Why? What am I getting wrong here?

FIRST SET OF VALUES:
gt value1="4" value2=":incline"
Select value="0.25"
gt value1="17" value2=":incline"
select value="0.25"
gt value1="18" value2=":incline"
select value="30"

SECOND SET IF VALUES:
gt value1="4" value2=":incline"
Select value="0.25"
gt value1="18" value2=":incline"
select value="30"

Zahnstocher:
> The following two sets of values SHOULD produce the same routing. but they don't. They produce different results. Why? What am I getting wrong here?

They should produce different results.

> FIRST SET OF VALUES:
> gt value1="4" value2=":incline"
> Select value="0.25"
> gt value1="17" value2=":incline"
> select value="0.25"
> gt value1="18" value2=":incline"
> select value="30"

Translation:
if 4 >= incline then select 0.25
else if 17 >= incline then select 0.25
else if 18 >= incline then select 30

Thus you can shrink it to:
if 17 >= incline then select 0.25
else if 18 >= incline then select 30

> SECOND SET IF VALUES:
> gt value1="4" value2=":incline"
> Select value="0.25"
> gt value1="18" value2=":incline"
> select value="30"

Translation:
if 4 >= incline then select 0.25
else if 18 >= incline then select 30

Obviously first and second set of values are not equal.
For example incline=10:

first set:
17 >= 10, thus 0.25 is selected.

second set:
4 is not >= 10, thus don't select 0.25.
So try next rule:
18 >= 10, thus 30 is selected.

"gt" means ">=" means "greater or equal".

Zahnstocher:
> >  gt value1="4" value2=":incline"
> > select value="6"
> > 
> > This means: If the incline is more than 4 per cent, multiply the height meters with 6.
> > So, if we have got a part of the route that has 100 meters elevation with 4 per cent slope, we multiply, and get a penalty of 600 seconds for that part. Right?
> 
> Right.

No, this is wrong.

It means:
if 4 >= incline then select 6.
You could also say:
if incline <= 4 then select 6.

Thus this is wrong:
> If the incline is more than 4 per cent, multiply the height meters with 6.

This is the correct version:
If the incline is **less or equal** than 4 per cent, multiply the height meters with 6.

To make it clear, look at this example:

gt value1="-3" value2=":incline" select value="1"
gt value1="3" value2=":incline" select value="2"
gt value1="7" value2=":incline" select value="4"
select value="15"

Means:
if -3 >= incline then select 1
else if 3 >= incline then select 2
else if 7 >= incline then select 4
else select 15

If this is easier to understand, you could rewrite it to:
if incline <= -3 then select 1
else if incline <= 3 then select 2
else if incline <= 7 then select 4
else select 15

Posteinwurf:
Dear Zahnstocher, 

thank you very much for the detailed explanation!
Now I get it: I have been reading the “greater than” the wrong way. When it meant if -3 >= incline, I have been reading if -3 =< incline

Thanks for your answer, now I can really start digging for appropriate values.

Malte

Posteinwurf:
I am currently trying to further refine the values. 

but the new version of osmand does not seem to to reognize the routing.xml 

have put it in net.osmand.plus and in the routing subfolder.

any ideas?

vshcherb:
Probably it is easy to use profiles (setup new profile in the main menu) there is a path for new routing.xml files

Posteinwurf:
thanks!

yes it says the path should be the routing subfolder. but I tried that. does not work.

Posteinwurf:
I have tried any possibility.

maybe the xml I am using has the wrong format? it worked with an older versipn of osmand. I downloaded it from 

https://github.com/osmandapp/OsmAnd-resources/tree/master/routing

that should be the current version, right?

Posteinwurf:
could someone try if osmand recognizes a modified routing.xml ?

Zahnstocher:
> could someone try if osmand recognizes a modified routing.xml ?

1. Download current routing.xml:
https://raw.githubusercontent.com/osmandapp/OsmAnd-resources/master/routing/routing.xml
2. Modify something (if you just want to test, if it works, you don't need to change anything)
3. Copy the routing.xml to the routing folder inside your osmand directory.
4. Einstellungen -> Anwendungsprofile -> + Hinzufügen -> Basisprofil z.B Fahrrad
Bei Profilname z.B. "test", **jetzt ganz wichtig, direkt darunter, Navigationsart anklicken und dort aus der Liste bicycle routing.xml auswählen.**
Für bessere Unterscheidbarkeit die Farbe anpassen, etwa auf "Grün".
5. Speichern

Jetzt kann bei der Navigation das grüne Fahrradsysmbol ausgewählt werden, um das neue Profil zu verwenden.

Posteinwurf:
many thanks!!!
--
Diese Nachricht wurde von meinem Mobiltelefon gesendet.Am 09.08.2019, 23:32, Zahnstocher <notifications@github.com> schrieb:


could someone try if osmand recognizes a modified routing.xml ?





Download current routing.xml:

https://raw.githubusercontent.com/osmandapp/OsmAnd-resources/master/routing/routing.xml

Modify something (if you just want to test, if it works, you don't need to change anything)

Copy the routing.xml to the routing folder inside your osmand directory.

Einstellungen -> Anwendungsprofile -> + Hinzufügen -> Basisprofil z.B Fahrrad

Bei Profilname z.B. "test", jetzt ganz wichtig, direkt darunter, Navigationsart anklicken und dort aus der Liste bicycle routing.xml auswählen.

Für bessere Unterscheidbarkeit die Farbe anpassen, etwa auf "Grün".

Speichern



Jetzt kann bei der Navigation das grüne Fahrradsysmbol ausgewählt werden, um das neue Profil zu verwenden.



—You are receiving this because you modified the open/close state.Reply to this email directly, view it on GitHub, or mute the thread.

[

{

"@context": "http://schema.org",

"@type": "EmailMessage",

"potentialAction": {

"@type": "ViewAction",

"target": "https://github.com/osmandapp/OsmAnd-resources/pull/444?email_source=notifications\u0026email_token=AEEWM4DWV7R5CI4GLK2ZCJ3QDXPATA5CNFSM4DTIVN62YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOD372MIY#issuecomment-520070691",

"url": "https://github.com/osmandapp/OsmAnd-resources/pull/444?email_source=notifications\u0026email_token=AEEWM4DWV7R5CI4GLK2ZCJ3QDXPATA5CNFSM4DTIVN62YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOD372MIY#issuecomment-520070691",

"name": "View Pull Request"

},

"description": "View this Pull Request on GitHub",

"publisher": {

"@type": "Organization",

"name": "GitHub",

"url": "https://github.com"

}

}

]

vshcherb:
You need to copy it in special folder "routing" , mentioned in configuration and yes we've tested it 4 weeks ago

rafaelilover1000-ctrl:
Depositar 5.000$ pesos a Toda la Gente  de Tijuana Baja California y dejar Este massage de parte de Rafael placido solo una ataque alas hacker en wilkipedia con el nombre Amelia ema Andrade y Gregorio Rio Rodriguez gonzalez y Erick ema Andrade by Adan ema Andrade y ulices Rodriguez gonzalez son estos tipos de ladrones atanquen Los en wilkipedia gracias.

rafaelilover1000-ctrl:
Develevop many  descargar

PR #446, opened 2017-09-09
Improvements to Finnish voice guidance

akikoskinen:
Check commit message for full details.

I transferred the file to a device and used the development tools in OsmAnd to listen through all the examples in the voice prompts section.

### DIFF
--- a/voice/fi/ttsconfig.p
+++ b/voice/fi/ttsconfig.p
@@ -30,7 +30,7 @@ string('distance.ogg', ', etäisyys ').
 
 % LEFT/RIGHT
 %string('prepare.ogg', 'Valmistaudu ').
-string('after.ogg', 'nach ').
+string('after.ogg', '').
 string('in.ogg', ' ').
 string('get.ogg', 'päästä  ').
 string('now.ogg', 'nyt,  ').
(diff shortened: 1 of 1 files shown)


voice/fi/ttsconfig.p
```diff
@@ -86,27 +86,26 @@ string('follow.ogg', 'Seuraa tietä ').
 
 % ARRIVE
 string('and_arrive_destination.ogg', 'ja olet perillä ').
-string('reached_destination.ogg', 'olet perillä ').
+string('reached_destination.ogg', 'olet perillä, ').
```
xmd5a2: Is this change really necessary?

akikoskinen: It's for adding a little pause before the destination name which can come after this. Without the comma it's just continuous: "olet perillä koti" ("koti" means home). With the comma, there's a slight pause: "olet perillä, koti". If there's another way to add that pause, then I'm willing to learn that.

akikoskinen: And the whole phrase just sounds more humane with that pause. Without the pause it sounds like a machine -- which of course it is :)

PR #447, opened 2017-09-16
Update routing.xml to avoid deprecated value

BalooUriza:
Don't look for deprecated "construction=yes" but instead "construction=minor"

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -72,7 +72,7 @@
 		<way attribute="access">
 			<select value="-1" t="osmand_change" v="delete"/>
 			<select value="-1" t="highway" v="construction"/>
-			<select value="-1" t="construction" v="yes"/>
+			<select value="-1" t="construction" v="minor"/>
 			<select value="-1" t="highway" v="motorway">
 				<if param="avoid_motorway"/>
 			</select>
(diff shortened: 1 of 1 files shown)


maraf24:
construction=minor is not replacement for construction=yes.
Road with construction=minor is, as wiki says, open and should not be excluded from routing.

BalooUriza:
construction=yes isn't a totally valid value, either, as also indicated.  While it should not be excluded, I do believe it should be given a lower weight than routes that do not have construction=minor if possible.

aceman444:
In that case it does not belong into the "access" block, but maybe in the "priority" block in routing.xml.
However it is hard to express something like "should be given a lower weight than routes that do not have construction=minor" in current syntax. You can only give such a road a single value and the order of the rules specify whether the highway=* wins or the construction=minor. You can't do multiplication of the values or something like that.

maraf24:
There is another problem with this patch.
construction=yes while deprecated, is still in use. And this is not Osmand role to enforce tagging according to the wiki. So this tag should be dropped only if it's no longer in use.

BalooUriza:
OK, perhaps I'm approaching this from the wrong prospective, then.  Perhaps osmand can treat construction=minor as highway=road (or some other basic road type) so that it is not necessarily the first route chosen?

PR #449, opened 2017-10-06
Adding names of mountain features

Ircama:
This PR adds the rendering support for the following mountain features:

- natural=arete
- natural=dale
- natural=couloir

It also revises:

- natural=valley
- natural=ridge

For *valley* and *ridge* (already available before), only the text label is now rendered, without any line.

This PR also includes:

- natural=mountain_range
- natural=massif
- place=region and region:type=mountain_area

For some unknown reasons, *natural=mountain_range*, *natural=massif* and *place=region / region:type=mountain_area* do not work even if the logic is the same as the other features. I checked these regions for instance:

- [natural=massif](https://www.openstreetmap.org/way/243593941)
- [natural=mountain_range](https://www.openstreetmap.org/relation/2698607)
- [place=region / region:type=mountain_area](https://www.openstreetmap.org/relation/2506537)

### DIFF
--- a/rendering_styles/default.render.xml
+++ b/rendering_styles/default.render.xml
@@ -5391,7 +5391,16 @@
 					<case minzoom="16" textSize="12" tag="natural" value="cave_entrance" textOrder="236"/>
 					<case minzoom="14" textSize="12" tag="natural" value="ridge" textOnPath="true" textDy="0" textOrder="77"/>
 					<case minzoom="16" textSize="12" tag="natural" value="cliff" textOnPath="true" textHaloRadius="3" textDy="0" textOrder="77"/>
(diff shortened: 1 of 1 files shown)


xmd5a2:
Added manually (partially) https://github.com/osmandapp/OsmAnd-resources/commit/11bbef771ed46832745652fcf739869b5f4c4e69

Ircama:
@xmd5a2 thanks, I would also suggest to consider arete, massif, mountain_range

Ircama:
@xmd5a2, very useful to check your development in [rendering_types.xml](https://raw.githubusercontent.com/osmandapp/OsmAnd-resources/master/obf_creation/rendering_types.xml) to also manage *arete*, *massif*, *mountain_range*, mapping them to *ridge* and *mountain_area* through:

    <entity_convert pattern="tag_transform" from_tag="..." from_value="..." to_tag1="..." to_value1="..."/>

Thanks again

PR #454, opened 2017-11-02
Add option to hide objects under construction

Ircama:
**Add option to hide objects under construction**

Default is to render objects under construction. This proposal introduces a new option which allows to hide them.

I’m not sure that the submitted modifications are enough to also have the label of the option automatically translated to all supported languages.

### DIFF
--- a/rendering_styles/default.render.xml
+++ b/rendering_styles/default.render.xml
@@ -111,6 +111,8 @@
 		type="boolean" possibleValues="" category="hide"/>
 	<renderingProperty attr="hideProposed" name="Hide proposed objects" description="Hide proposed objects"
 		type="boolean" possibleValues="" category="hide"/>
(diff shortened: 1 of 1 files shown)


xmd5a2:
Why hide objects under construction?

Ircama:
I was considering that features under construction, especially roads (*highway* elements), even if relevant for future scope, might be misleading and sometimes also could clog the map with information not useful for the present.

The option to hide them will simply support people just interested in current road conditions.

Anyway, this PR is just a proof of concept; if the proposed option is not generally considered useful, the PR can be dropped/closed without merging.

maraf24:
>The option to hide them will simply support people just interested in current road conditions.

highway=construction is current road condition and it is useful for people because it usually effects neighbouring roads, which is not always properly mapped.

Ircama:
> highway=construction is current road condition and it is useful for people because it usually effects neighbouring roads, which is not always properly mapped.

OK, let's put this PR in standby.

PR #457, opened 2017-11-04
Improve visibility of mountain huts if moreDetailed="true"

Ircama:
**Improve visibility of mountain huts if moreDetailed="true"**

Mountain huts are reference buildings within undeveloped zones (typically in the mountains) and should be rendered at wide zooms.

With this PR, *alpine_hut* and *wilderness_hut* will be most probably rendered at z14+ or z15+ (before, some *alpine_hut* like [Rifugio Rino Olmo]( https://www.openstreetmap.org/way/120656176) was rendered at z18+).

This proposal is based on the comparison of Osmand with classic 1:50000 – 1:25000 touristic and topographic paper maps, that usually show these features.

I tried to reuse a similar design pattern to [Tweak place_of_worship order](https://github.com/osmandapp/OsmAnd-resources/commit/10184e1501f759724d184f1ab01f5b35fe7f52e4#diff-aa68e561bc04596c485828a0f39f460c).

### DIFF
--- a/rendering_styles/default.render.xml
+++ b/rendering_styles/default.render.xml
@@ -2545,6 +2545,10 @@
 				<case tag="attraction" value="train"/>
 				<case tag="attraction" value="log_flume"/>
 				<case tag="building" value="no" order="-1"/>
+				<switch moreDetailed="true" order="10" ignorePolygonArea="true" ignorePolygonAsPointArea="true">
+					<case tag="tourism" value="alpine_hut"/>
+					<case tag="tourism" value="wilderness_hut"/>
+				</switch>
 				<case tag="amenity" value="place_of_worship" ignorePolygonAsPointArea="true">
 					<apply_if moreDetailed="true" ignorePolygonArea="true"/>
 				</case>
@@ -5884,6 +5888,7 @@
 					<case minzoom="15" tag="tourism" value="lean_to"/>
 					<case minzoom="14" tag="tourism" value="wilderness_hut"/>
 					<case minzoom="15" tag="amenity" value="shelter"/>
+					<apply_if minzoom="15" moreDetailed="true" textOrder="43"/>
 					<apply_if engine_v1="true" textDy="5"/>
 					<apply_if nightMode="true" textHaloColor="#55234ca7"/>
 				</switch>
@@ -6802,6 +6807,7 @@
 						<case tag="tourism" value="lean_to" icon="lean_to"/>
 						<case tag="tourism" value="wilderness_hut" icon="wilderness_hut"/>
 						<apply shield="blue_round_shield"/>
+						<apply_if minzoom="14" moreDetailed="true" iconOrder="43" iconVisibleSize="9" intersectionSizeFactor="0.7"/>
 						<apply_if minzoom="15" additional="drinking_water=yes" icon_2="drinking_water_yes_map"/>
 						<apply_if minzoom="15" additional="drinking_water=no" icon_2="drinking_water_no_map"/>
 						<apply_if nightMode="true" shield="blue_round_night_shield"/>
@@ -7293,6 +7299,7 @@
 						<case minzoom="13" tag="tourism" value="wilderness_hut" icon="wilderness_hut"/>
 						<case minzoom="16" tag="amenity" value="shelter" icon="amenity_shelter"/>
 						<case minzoom="15" tag="leisure" value="summer_camp" icon="summer_camp"/>
+						<apply_if minzoom="14" moreDetailed="true" iconOrder="43" iconVisibleSize="9" intersectionSizeFactor="0.7"/>
 						<apply_if minzoom="16" additional="drinking_water=yes" icon_2="drinking_water_yes_map"/>
 						<apply_if minzoom="16" additional="drinking_water=no" icon_2="drinking_water_no_map"/>
 					</switch>


Ircama:
With this additional note, I wish to provide an example of where this PR is appropriate for Osmand and why it also needs to be complemented.

OpenStreetMap renders *alpine_hut* and *wildernes_hut* icons (but not text) at [z13+](https://www.openstreetmap.org/way/120595368#map=13/45.9421/10.0770). At [z14+](https://www.openstreetmap.org/way/120595368#map=14/45.9421/10.0770) also text is shown. No icon and text at [z12](https://www.openstreetmap.org/way/120595368#map=12/45.9422/10.0772). Reason for such rendering (which is appropriate IMHO) follows [this piece of code for icons](https://github.com/gravitystorm/openstreetmap-carto/blob/master/amenity-points.mss#L37-L51) and [this one for texts](https://github.com/gravitystorm/openstreetmap-carto/blob/master/amenity-points.mss#L1749-L1775).

Understanding how Osmand renders features at various zooms in definitely not easy as many style properties are unexplained and the code is not immediate to unravel. There is a difference if the feature is a point or a polyline. Before this PR, Osmand often rendered polyline buildings with *alpine_hut/wilderness_hut* attributes at much more detailed zooms, like z19+.

z18 olmo  z14 olmo n
Rino Olmo Refuge, z18 before, z14 after (not rendered)

z19 olmo  z15 olmo n
Rino Olmo Refuge, z19 before, z15 after (rendered)

z18 cassinelli  z14 cassinelli n
Cassinelli Refuge, z18 before, z14 after (not rendered)

z19 cassinelli  z15 cassinelli n
Cassinelli Refuge, z19 before, z15 after (rendered)

The reason why also after applying this PR features are still not shown at z14 is unknown and IMHO this needs a further update from a more experienced developer.

I meanwhile rebased the PR.

rendering_styles/default.render.xml
```diff
@@ -2545,6 +2545,10 @@
 				<case tag="attraction" value="train"/>
 				<case tag="attraction" value="log_flume"/>
 				<case tag="building" value="no" order="-1"/>
+				<switch moreDetailed="true" order="11" ignorePolygonArea="true" ignorePolygonAsPointArea="true">
```
xmd5a2: Normally polygon order should be in range 1-10. order="11" means that polygon will cover some ways.

Ircama: OK, thanks so much for this clarification.

I'll correct this to 10 in a few hours and will do again some testing, if this new value is also fine to you.

Why the polygon is not rendered at z14?

xmd5a2: Don't know. I need exact place location to test it.

Ircama: Rino Olmo Refuge: https://www.openstreetmap.org/way/120656176
Cassinelli Refuge: http://www.openstreetmap.org/way/120595368

xmd5a2: These objects were filtered by area during obf creation process and there is no easy way to fix it.

Ircama: Ah, I did not realize about the existence of such filter, I'll have a look at the code.

I thought that

    java -Djava.util.logging.config.file=loging.properties -Xms64M -Xmx512M -cp "./OsmAndMapCreator.jar;./lib/OsmAnd-core.jar;./lib/*.jar" net.osmand.binary.BinaryInspector -vmap -vmapobjects -bbox=10.0219093238,45.9338562315,10.0323629379,45.285758321 -zoom=10 Italy_lombardia_europe_2.obf

was just failing for some other reason (z10-), as strangely no feature at all is returned at that specific zoom (working fine instead at z11+).

This is the output at z10:
```
Binary index Italy_lombardia_europe_2.obf version = 2 edition = Wed Nov 01 01:00:00 CET 2017
1 Map data Italy_lombardia_europe - 195590432 bytes
        1.1 Map level minZoom = 15, maxZoom = 22, size = 119832438 bytes
                Bounds (left top - right bottom) : 5.0456, 48.4093 NE - 16.5914, 43.4945 NE
        1.2 Map level minZoom = 13, maxZoom = 14, size = 45800063 bytes
                Bounds (left top - right bottom) : 5.0456, 48.4093 NE - 16.5914, 43.4945 NE
        1.3 Map level minZoom = 12, maxZoom = 12, size = 21744576 bytes
                Bounds (left top - right bottom) : 5.0456, 48.4093 NE - 16.5914, 43.4945 NE
        1.4 Map level minZoom = 11, maxZoom = 11, size = 8178247 bytes
                Bounds (left top - right bottom) : 5.0456, 48.4092 NE - 16.5914, 43.4945 NE
        Total map objects: 0
2 Routing data Italy_lombardia_europe - 55347817 bytes
        Bounds (left top - right bottom) : 8.2653, 48.4093 NE - 16.5638, 44.5074 NE
3 Address data Italy_lombardia_europe - 25309012 bytes
        3.1 Address part size=8434728 bytes
        3.3 Address part size=9469239 bytes
        3.2 Address part size=2618490 bytes
4 Poi data Italy_lombardia_europe - 32302645 bytes
5 Transport data Italy_lombardia_europe - 1971393 bytes
        Bounds (left top - right bottom) : 8.7088, 46.5545 NE - 11.1244, 44.9818 NE
```

Does this mean that all features related to this bounding box are all filtered during obf creation for that specific zoom?

xmd5a2: Detailed maps contains no data at zoom 10. Look at the default map zoom levels in osmand map creator settings: 11;12;13-14;15-

Ircama: @xmd5a2 thanks again, now it is clear.

Ircama:
> These objects were filtered by area during obf creation process and there is no easy way to fix it.

OK, I would then suggest to keep this PR the way it is now. I corrected the *order* as per request.

rendering_styles/default.render.xml
```diff
@@ -6802,6 +6807,7 @@
 						<case tag="tourism" value="lean_to" icon="lean_to"/>
 						<case tag="tourism" value="wilderness_hut" icon="wilderness_hut"/>
 						<apply shield="blue_round_shield"/>
+						<apply_if minzoom="14" moreDetailed="true" iconOrder="43" iconVisibleSize="9" intersectionSizeFactor="0.7"/>
```
xmd5a2: What was the purpose of this change? Why are you lowering iconOrder starting from zoom 14? Objects with iconOrder="43" will be rendered over iconOrder="81". Did you mean maxzoom="14" ?

Ircama: This is indeed wanted. I am lowering iconOrder at z14+ (and not z14- !) to ensure that mountain huts will have better priority to be rendered over other features starting from zoom 14, in case moreDetailed is also checked in Osmand settings.

PR #460, opened 2017-11-16
Documentation in Routing.xml

pietervdvn:
The added documentation is based on educated guesses. If possible, read it once to see if no glaring errors are in there.

Some more documentation about numeric parameters, srtm-data and comparing keys should be needed. If some maintainer would be able to give a few pointers, I'll add those in as well.

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -341,27 +341,6 @@
 	The `type` gives a hint on how to parse the values. If the type is `height`, '1' will be interpreted as '1 meter'.
 	For a type `height`, '1' will be interpreted as one ton (1000 kg)
 
-	### Using if/gt tags in combination with select
-
-	A different and equivalent format for if- and gt-tags is placing an if inside a select. However, it is deprecated (so don't use it)
-
-	<if param="$x">
-		<select value="..."/>
-	</if>
-
-	is totally equivalent to:
-
-	<select value="...">
-		<if param="$x"/> (Deprecated notation)
-	</select>.
-
-	However, we recommend *not to use this format*. The if as outer tags are clearer, and support multiple selects:
-
-	<if param="$x">
-		<select value="1" t="y" v="z"/>
-		<select value="0" t="w" v="a"/>
-		...
-	</if>
 
        
     Miscellaneous
@@ -371,6 +350,30 @@
 	osmand_highway_integrity_brouting represents how good a way is for a bicycle.
 	Mostly it merges smoothness, surface and tracktype into one value. It is calculated internally?
 
+	### Alternative notation for <if><select /></if> and <gt><select /></gt>
+
+	A different but equivalent format for if- and gt-tags exists: the <if> can be placed inside a <select>
+	However, it is considered bad style (so don't use it).
+
+	In other words:
+
+	<if param="x">
+		<select value="..."/>
+	</if>
+
+	is totally equivalent to:
+
+	<select value="...">
+		<if param="x"/> (Deprecated notation)
+	</select>.
+
+	However, we recommend *not to use this format*. The if as outer tags are clearer, and support multiple selects:
+
+	<if param="$x">
+		<select value="1" t="y" v="z"/>
+		<select value="0" t="w" v="a"/>
+		...
+	</if>
 
 
 	-->


routing/routing.xml
```diff
@@ -23,29 +23,226 @@
...
-                <if param="parameter id"/>
-              </select>
+	
+		If you want to tweak the routing engine, read this carefully.
+		
+    Deploying your own routing.xml
+    ==============================
+		
+		When you finish editing this document (or want to test on your Android device), copy this file over to:
+		sdcard/Android/data/net.osmand.plus/files/
```
xmd5a2: This path may differ on different versions and depending on the setting "Data storage folder"

xmd5a2: Also iOS version of OsmAnd exists, however it's not easy to use custom routing.xml

routing/routing.xml
```diff
@@ -23,29 +23,226 @@
...
+		A profile block contains at least the following segments:
+		1) access                 (which determines if a road can be taken)
+   		2) speed                  (which determines what max speed a road allows)
+    	
+    	A profile block might also contain following segments:
+    	3) oneway                 (which determines in which direction a road can be taken)
+    	4) priority               (which determines how nice the road is to drive on)
+    	5) obstacle               (which determines the delay of an obstacle, in seconds for the routing engine)	
+    	6) obstacle_time          (which determines the delay of an obstacle, in seconds for the human)
+    	7) penalty_transition     (which determines the penalty of transitioning from one road surface to another)
```
xmd5a2: That's not true. penalty_transition is penalty for the exit from a high-level road to a low-level road (primary to primary_link e.g.)

routing/routing.xml
```diff
@@ -23,29 +23,234 @@
...
+        In other words, the values here do not influence routing.
+        
+        
+
+    ## 7) Penalty_transition (specified on ways, penalty time in seconds)
+
+        When driving a car, changing from a motorway to a trunk slows you down: you should already slow down 500 meters before the trunk.
+        This has a penalty which is taken into account by (only) the navigation algorithm.
+        This is done by taking the difference between two penalties.
+        The formula used is `minimum(0, penalty_new_way - penalty_current_way)`.
```
Zahnstocher: maximum

routing/routing.xml
```diff
@@ -23,29 +23,234 @@
...
+        E.g. 
+        <way attribute="penalty_transition">
+            <select value="0" t="highway" v="motorway"/>
+            <select value="5" t="highway" v="motorway_link"/>
+            <select value="20" t="highway" v="tertiary"/>
+        </way>
+        
+        
+        This means that, going from a `motorway_link` to a `tertiary`, the penalty given will be
+        `minimum(0, 20 - 5)` or `15 seconds`.
```
Zahnstocher: maximum

routing/routing.xml
```diff
@@ -23,29 +23,234 @@
...
+        
+        On the other hand, going from motorway to motorway incurs no penalty, neither does going from a motorway_link to a motorway.
+
+    ## 8) obstacle_srtm_alt_speed (specified on ways, in km/h)
+    
+        This block is used to give alternative speeds, based on the incline of a road, based on BRouter.
```
Zahnstocher: This has nothing to do with BRouter.

routing/routing.xml
```diff
@@ -23,29 +23,234 @@
...
+            <select value="20" t="highway" v="tertiary"/>
+        </way>
+        
+        
+        This means that, going from a `motorway_link` to a `tertiary`, the penalty given will be
+        `minimum(0, 20 - 5)` or `15 seconds`.
+        
+        On the other hand, going from motorway to motorway incurs no penalty, neither does going from a motorway_link to a motorway.
+
+    ## 8) obstacle_srtm_alt_speed (specified on ways, in km/h)
```
Zahnstocher: These values are not alternative speeds measured in km/h, higher values mean higher penalty, not higher speed.

routing/routing.xml
```diff
@@ -23,29 +23,234 @@
...
+        <select value="1" t="oneway" v="1"/>
+        
+        And if, for some reason, you want to go against the oneway, the value of "-1" should be specified:
+        
+        <select value="-1" t="oneway" v="-1"/>
+            
+        
+    ## 4) Priority (specified on ways, between 0 and 1)
+    
+        Priority is a preference multiplicator (between 0 and 1), which can only be used to _de_prioritize routes.
```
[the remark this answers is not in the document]
pietervdvn: I know, I know... I find them to be highly confusing. According to [https://osmand.net/build_it] (see title 'Customizing routing engine' at the bottom of the page), it should be between 0 and 1.

In my limited experience, values >1 didn't do much as well...

Could you point me to the source code (preferable the 'safe' java versions - I'm not proficient in reading C++) where the documents are parsed and used? The source can't lie

routing/routing.xml
pietervdvn: Yes, I often mix those up...

routing/routing.xml
pietervdvn: I have no idea what this block does. Please point me to the docs, source code or just explain.

pietervdvn:
I've updated the documentation again. Please review it further and point out missing and faulty parts.

routing/routing.xml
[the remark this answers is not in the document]
Zahnstocher: defineSpeedPriority():
https://github.com/osmandapp/Osmand/blob/7113eb554fe7af9f4bb5257ccae0258d22c60e49/OsmAnd-java/src/net/osmand/router/GeneralRouter.java#L360
Default prio is 1.

calculateTimeWithObstacles():
https://github.com/osmandapp/Osmand/blob/47b4d762a7e4b27388619a9bf2ab1a279d4c0fa1/OsmAnd-java/src/net/osmand/router/BinaryRoutePlanner.java#L581

speed_routing = minimum(speed * prio, maxDefaultSpeed)

Thus it depends on speed and maxDefaultSpeed how big prio can be, until it has no additional effect anymore.

Example:
bicycle
maxDefaultSpeed="33"
residential speed 23
Thus prio can be 23/33 = 1.43... for residential road, if it is bigger it is limited to 1.43..., for example prio = 2 will only have the effect of prio 1.43...

routing/routing.xml
```diff
@@ -23,29 +23,238 @@
...
+        
+        
+        This means that, going from a `motorway_link` to a `tertiary`, the penalty given will be
+        `maximum(0, 20 - 5)` or `15 seconds`.
+        
+        On the other hand, going from motorway to motorway incurs no penalty, neither does going from a motorway_link to a motorway.
+
+    ## 8) obstacle_srtm_alt_speed (specified on ways, in seconds of penalty)
+    
+	What the fuck does this block?
```
Zahnstocher: Meaning of select value:
penalty in seconds for every height metre

For example:

>				<gt value1="-60" value2=":incline">
>					<select value="35"/>
>				</gt>
>				<gt value1="-25" value2=":incline">
>					<select value="30"/>
>				</gt>
>				<gt value1="-17" value2=":incline">
>					<select value="17"/>
>				</gt>
>				<gt value1="-9" value2=":incline">
>					<select value="5"/>
>				</gt>
>				<gt value1="-3" value2=":incline">
>					<select value="1"/>
>				</gt>
>				<gt value1="3" value2=":incline">
>					<select value="1"/>
>				</gt>
>				<gt value1="7" value2=":incline">
>					<select value="4"/>
>				</gt>
>				<gt value1="13" value2=":incline">
>					<select value="8"/>
>				</gt>
>				<gt value1="25" value2=":incline">
>					<select value="10"/>
>				</gt>
>				<select value="15"/>

has the following meaning:

more than 25% => 15 seconds per height metre
[25% 13%) => 10 seconds per height metre
[13% 7%) => 8 seconds per height metre
[7% 3%) => 4 seconds per height metre
[3% -3%) => 1 seconds per height metre
[-3% -9%) => 1 seconds per height metre
[-9% -17%) => 5 seconds per height metre
[-17% -25%) => 17 seconds per height metre
[25% -60%) => 30 seconds per height metre
60% and less => 35 seconds per height metre

negative select value (seconds per height metre) means "impassable"

Let the incline of a part of a way be 20%, then the penalty per height metre is 10 seconds.
If the part of the way has 9 height meters, the penalty for this part will be 10*9 = 90 seconds.

The height penalty is additional.
Total routing engine time is:
height penalty + distance/(speed*prio) + obstacles + turn penalties + ....

pietervdvn:
I'm a bit confused about the usage of `:` and `$` in `<gt value=` tags.

Is there a difference between both of them? I assume that `:key` substitutes to the value of the way, whereas `$param` is substituted by a numeric parameter?

But that is contradicted by a statement as the following:

			<select value="-1" t="maxheight">
				<gt value1=":height" value2="$maxheight" type="length"/>
			</select>


Furthermore, what does the `type="length"`-part do? And, is this equivalent with following snippet? Also in the absence of a maxheight-key?

        <gt value1=":height" value2="$maxheight">
                <select value="-1" t="maxheight">
         </gt>

routing/routing.xml
pietervdvn: I copied your explanation and cleaned it up a little. New commit coming soon.

routing/routing.xml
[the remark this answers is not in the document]
pietervdvn: Added to the docs! Thanks for the explanation.

Is this value only for the routing engine? And is only speed used to calculate the ETA on the UI?

Zahnstocher:
> I'm a bit confused about the usage of `:` and `$` in `<gt value=` tags.
> 
> Is there a difference between both of them? I assume that `:key` substitutes to the value of the way, whereas `$param` is substituted by a numeric parameter?
> 
> But that is contradicted by a statement as the following:
> ```
> 		<select value="-1" t="maxheight">
> 			<gt value1=":height" value2="$maxheight" type="length"/>
> 		</select>
> ```

Yes, this is very confusing, because it is not used consistently.
":incline" is the value which is inside the OsmAnd obf file.
But ":height" is the parameter which is set by the user (the height of users vehicle) and $maxheight is the value which is inside the OsmAnd obf file (value of the OSM way).

See: https://github.com/osmandapp/Osmand/blob/45003c897d2a456042afba4b831ce5d729578744/OsmAnd-java/src/net/osmand/router/GeneralRouter.java

> Furthermore, what does the `type="length"`-part do? And, is this equivalent with following snippet? Also in the absence of a maxheight-key?
> 
> ```
>     <gt value1=":height" value2="$maxheight">
>             <select value="-1" t="maxheight">
>      </gt>
> ```

It is not the same.
The type determines, how the value is parsed.
Type length could be km, miles, ...: parseLength()
Type weight can be kg, lbs, ...: parseWeightInTon()
See: https://github.com/osmandapp/Osmand/blob/508054cfec1cc4c43d87aa475b9932433c734694/OsmAnd-java/src/net/osmand/binary/RouteDataObject.java

routing/routing.xml
```diff
@@ -23,29 +23,336 @@
...
+  			<select value="1"/>
+  		</gt>
+  		<gt value1="3" value2=":incline">
+  			<select value="0"/>
+  		</gt>
+
+  		<gt value1="17" value2=":incline">
+  			<select value="1"/>
+  		</gt>
+  		<gt value1="13" value2=":incline">
```
Zahnstocher: It does not make much sense to have 13 after 17, because 17 will always match before 13, so that 13 is not reachable.

pietervdvn:
Thanks for replying again to this pull request. I'm having very busy weeks ahead, so I can't look to this in detail right now. I will however work further on this in 2 or 3 weeks.

routing/routing.xml
pietervdvn: Good catch, fixed

pietervdvn:
Hi!

Finally found a bit of time to fix this further up.

About `:` vs `$`: I picked an arbitrary idiom (`:` for parameters of the file, `$` for OBF-data) and updated routing to be consistent.

pietervdvn:
At last: are there differences between:

    <if param='x'>
        <select value="-1"  t="key" v="someValue"/>
    </if>


    <select value="-1"  t="key" v="someValue"/>
        <if param='x'>
    </select>

aceman444:
I think the former version is better because you can then put multiple key/value pairs inside the 'if'. If for single selects both versions work, then the former could be chosen again for consistency.

pietervdvn:
@aceman444 You are definitely right about that. There are 13 occurrences of the latter form - but I would first like to get a confirmation that both are effectively the same.

So: if both are the same, then I'll change those and document this. Then I would assume the documentation process is completed and we can finally merge the docs in the trunk.

vshcherb:
They are effectively the same and being reduced into 2nd form. 
The way calculation works (may be it is worth to mention):
1. Each object has  tags 
2. Each tag/value pair (!) is transformed into a number, so each object (road) is associated with bitset [1, 5, 10]. 
3. The section is also an array of integers [11, 2, 5 (if ...), 1 (if ...),  ... ]. 
4. Iterating through second array and checking if the tag/value present and then evaluating if gives a number.

It already looks that quite a lot of work done, is it worth to merge it as it is ? 
@pietervdvn @Zahnstocher @sonora

pietervdvn:
Uniformisation has happened, the alternate form is documented.

I didn't really get your explanation of the arrays. However, I don't think that the internal workings of the routing algorithm are relevant for someone who wants to tweak his routing - so IMHO I wouldn't add this into the documentation.

As far as I am concerned, this PR can be merged and closed.

vshcherb:
Something wrong with this pull request may be it should be reimplemented
+17,514 −2,051  this is too many changes and lots of them in binary files...

pietervdvn:
Oops, that is indeed not what should happen. Probably something went wrong while merging/rebasing.

I have forked the repo again and updated only the routing.xml file. I've opened a new pull request, #487 . Please merge that one instead and, if needed, continue discussions there.

PR #462, opened 2017-11-29
Update country specific living_street access

hakuchi:
With this update all routing profiles from OsmAnd should be supported.

### DIFF
--- a/obf_creation/rendering_types.xml
+++ b/obf_creation/rendering_types.xml
@@ -1551,8 +1551,9 @@
 		<type tag="highway" value="path" minzoom="12" nameTags="ref"/>
 		<type tag="highway" value="living_street" minzoom="12" nameTags="ref"/>
 		<!-- According to https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access-Restrictions -->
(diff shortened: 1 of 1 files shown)


xmd5a2:
What's the point of this pull request? According to https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access-Restrictions#Russia hgv+living_street=destination in Russia. Your pull request removes this rule.

hakuchi:
[#4662](https://github.com/osmandapp/Osmand/issues/4662) and comments on this https://github.com/osmandapp/OsmAnd-resources/commit/5503e077d951be3145717773d9ec6e26f2b607e1 I think I can fix with this fix.

I dont think that my request remove the hgv from this rule, because in the [1556 line](https://github.com/osmandapp/OsmAnd-resources/pull/462/commits/6616c13d9b3027e823b0961252e658136db45300#diff-9f147a86fcc6b9151b17cf840bf4a500R1556) I have set Russia and Ukraine to access vehicle=destination(any vehicle)

In the first line the 5 countries have access motor_vehicle=destination(all vehicles with a motor)

PR #463, opened 2017-11-29
Show text label for natural=rock

Ircama:
**Show text label for *natural=rock* in case of ways/polygons (and not only for nodes).**

As *natural=rock* is converted to *natural=bare_rock* for "way" in *rendering_types.xml*, the text label for *natural=bare_rock* is added.

Notice that this change does not yet add "natural_stone" icon to "bare_rock", just assuming that this is really wanted (otherwise,  `<case minzoom="17" tag="natural" value="bare_rock"/>` shall also be added before `<apply icon="natural_stone" shield="stone_shield"/>`).


xmd5a2:
Many natural=rock tags are actually natural=bare_rock http://overpass-turbo.eu/s/txC so I think we shouldn't add icons to polygonal bare_rock.

Ircama:
> Many natural=rock tags are actually natural=bare_rock http://overpass-turbo.eu/s/txC so I think we shouldn't add icons to polygonal bare_rock.

OK, right!

PR #471, opened 2017-12-07
changing "if_starts_with_value1" road shields from Italy.

hakuchi:
I have some problems with the road shields in Italy.
Some road shields are displayed correct and some **only in white**.
I can not find anything wrong in the OSM data.
Tested with latest map and nightly.

I don't know if this request will fix the issue.
I think that the obf creation process is not detecting the correct letter because of a lot of "S" in Italian ref.

I deleted "SS", "SR", "SP" and added only "s" in `if_starts_with_value1="s"`
Also I deleted "LS/SP" and added only "L" in `if_starts_with_value1="l"`
because I think this work better with the obf creation process.
     
     
here some screenshots around [Meran](https://www.openstreetmap.org/#map=15/46.6676/11.1648)
same ref with different color:

2 color
SS and LS/SP with wrong color:

SS SP


xmd5a2:
Looks like you have a mix of very old and new maps because I don't experience this issue
screenshot from 2017-12-12 12-06-19

hakuchi:
Thx for your check.
Strange...I made a new clean installation of OsmAnd nightly and ref is displayed correct.

PR #473, opened 2017-12-09
add voice logic for very short distances (<17m/yards)

tobydickenson:
Change for osmandapp/Osmand#4766.

This introduces voice guidance with 1m or 1 yard precision for very short distances.

I have tested this in en-gb, and made corresponding edits to all other languages with the exception of ru (Russian) and ro (Romanian) which already had some logic involving short distances.


aceman444:
Thanks for doing all the other languages too!

PR #483, opened 2018-02-06
Update routing.xml

Posteinwurf:
I would suggest to move the footway=sidewalk further down, because there were cases when a cycleway was marked with this tag, like in this case: https://github.com/osmandapp/Osmand/issues/4461 With the change, Osmand will first check if it is cycleway=track (etc), and only downgrade sidewalks, if not cycleway tag was found.

I have also moved segregated=yes further down.

Additionally, I would suggest to downgrade mountainbike tracks. Even if mountainbike enthousiasts think that a way gets 3 points (for mountainbiking), it might still be terrible for normal cycling! Those mountainbike tags are quite subjective anyway (do the easy tracks get many points or the horribly difficult ones? That is not clear at all. So, I would treat any MTB track with caution in a general bicycle router).

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -853,10 +853,8 @@
 <!--				<select value="0.6" t="surface" v="wood"/>-->
 			</if>
 			
-			<select value="1.0" t="segregated" v="yes"/>
 			<select value="1.0" t="bicycle" v="dismount"/>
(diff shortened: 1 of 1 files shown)


vshcherb:
Looks good to me.

PR #484, opened 2018-02-12
Fixed issues with Czech Google TTS by using colons

StenSoft:
- prepositions read as initials after turn
    “odbočte na K Vozovně“ was read as “odbočte na K. Vozovně”

- time estimate read in genitive instead of nominative
    “potřebná doba 45 minut“ was read as “potřebná doba čtyřiceti pěti minut”


StenSoft:
I have received report that it may not work correctly with street names that actually begin with an initial. I will need to test this with such a name before reopening.

StenSoft:
Tested and works correctly

PR #487, opened 2018-03-09
Added documentation to routing.xml and made everything more consistent

pietervdvn:
This PR replaces #460, where something went wrong when rebasing.

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -23,30 +23,362 @@
 
 	<!-- HELP INFORMATION ABOUT FILE -->
 	<!--
-	   Description of blocks:
-	     1) <way attribute="access">
-	        Defines acceptable roads for routingProfile. Value of -1 means way impassable,
-	        value of 1 means way passable.
-	     2) <way attribute="speed" type="speed">
-	        Defines speed limits on roads in km/h (if it is not specified on the road via maxspeed tag)
-	     3) <way attribute="priority">
-	     	Priority is a preference multiplicator for the road type (consider it is an accelerator of the road)
-	     4) <point attribute="obstacle">
-	        Obstacle defines a point of the road and whether it can be passed by the vehicle.
-	        Penalty is measured in seconds, a value of -1 means the obstacle cannot be passed and a value of 0 means "no penalty"!
-	        For routing engine.
-             5) <point attribute="obstacle_time">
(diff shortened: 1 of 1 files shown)


routing/routing.xml
```diff
@@ -740,74 +1074,74 @@
 
 		<way attribute="obstacle_srtm_alt_speed">
 			<if param="height_obstacles">
-				<gt value1="-100" value2=":incline">
```
vshcherb: Is it correct change from semicolon to dollar-sign? It has different semantics

pietervdvn: Bit of history: I did a PR earlier where I fucked up some binary files, so I created a cleaner version.

There was a lot of discussion about all these details in #460. One of them is [this comment](https://github.com/osmandapp/OsmAnd-resources/pull/460#issuecomment-363588384), which states that they are the same.

I could be mistaken of course, in that case, I'll gladly here what the difference in semantics are.

vshcherb: I don't understand where did you get there is no difference:
- $maxweight - is the value of the tag "maxweight" of 
- : incline is a routing parameter and it is specified in routing settings
-<gt value1=":height" value2="$maxheight" type="length">

pietervdvn:
The linked comment says:

> ":incline" is the value which is inside the OsmAnd obf file.
But ":height" is the parameter which is set by the user (the height of users vehicle) and $maxheight is the value which is inside the OsmAnd obf file (value of the OSM way).

This implies that: `$maxheight` and `:incline`are values which are both read from the .obf-file, thus implying there is no difference.

routing/routing.xml
pietervdvn: No, there is no difference. Please see [this comment](https://github.com/osmandapp/OsmAnd-resources/pull/460#issuecomment-363588384)

pietervdvn:
Once more, I would like to vow to get this documentation merged. It will benefit us all, especially OsmAnd. BTW, I'm planning a small meetup with the community where we will be tweaking our personal route planners. It would be cool if it was merged by then.

vshcherb:
I think there is a misunderstanding in this PR and it is not going to be merged just because it consists of changes that are not related only to Documentation. 
Again I think the change is not correct and it will break the routing compatibility. Anyway it is better not to change anything logical if the PR is called "Update documentation", so I would advise to split this PR in 2 parts.

PR #488, opened 2018-03-15
Fix some grammar in pl voice

GITNE:
e.g. “ograniczenie prędkości 50” was read as “ograniczenie prędkości pięćdziesięciu”.

There is still some bad grammer left, especially when reading `ref` or `destiantion:ref` values in specific contexts. Additionally, street names are not read in the accusative form in turn commands.

### DIFF
--- a/voice/pl/config.p
+++ b/voice/pl/config.p
@@ -93,7 +93,7 @@ string('reached_poi.ogg', 'mijasz POI ').
 
 % ATTENTION
 %string('exceed_limit.ogg', 'przekraczasz dozwoloną prędkość ').
-string('exceed_limit.ogg', 'ograniczenie prędkości ').
+string('exceed_limit.ogg', 'ograniczenie prędkości do ').
 string('attention.ogg', 'uwaga, ').
 string('speed_camera.ogg', 'fotoradar ').
 string('border_control.ogg', 'kontrola graniczna ').
--- a/voice/pl/ttsconfig.p
+++ b/voice/pl/ttsconfig.p
@@ -92,7 +92,7 @@ string('reached_poi.ogg', 'mijasz POI ').
 
 % ATTENTION
 %string('exceed_limit.ogg', 'przekraczasz dozwoloną prędkość ').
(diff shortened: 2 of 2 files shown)


GITNE:
FYI: I have just realized why TTS reads e.g. `ograniczenie prędkości pięćdziesięciu` instead of `ograniczenie prędkości piećdziesiąt`, like one would assume in a more literal translation from English—if this is what one actually wants. It is probably because of the full stop `. U+002E FULL STOP` (or period) at the end of the complete voice command. After reviewing #484, the later form could probably be achieved by adding a colon before the number and a space after the number, e.g. `ograniczenie prędkości: 50 `(`.`). However, I think that `ograniczenie prędkości do pięcdziesięciu` sounds more natural—while `ograniczenie prędkości pięćdziesięciu` was definitely incorrect—and most native speakers will probably prefer a more naturally sounding turn-by-turn navigation over a literal translation from English while driving/biking etc.

maraf24:
> e.g. “ograniczenie prędkości 50” was read as “ograniczenie prędkości pięćdziesięciu”

String  produced by ttsconfig.pl is just "ograniczenie prędkości 50", so it depends on tts engine how the number "50" is translated.
With your changes, less sophisticated engines would now generate "ograniczenie prędkości do pięćdziesiąt". But I suppose no one is using such engines nowadays:)

GITNE:
@maraf24 Thank you for the clarification. I have been testing and referring to the AOSP TTS. For better or worse, it is the only TTS you can more or less safely assume and develop for. So, I guess this change is okay.

Anyhow, given your information I wonder now why the TTS reads “pięćdziesięciu” instead of “pięćdziesiąt”. Do you have any idea?

GITNE:
> Anyhow, given your information I wonder now why the TTS reads “pięćdziesięciu” instead of “pięćdziesiąt”. Do you have any idea?

Hmm, the TTS may still always add a full stop at the end, regardless of what text it has been passed. You would probably have to check that code in AOSP to know for sure…

PR #496, opened 2018-07-23
Added documentation to routing.xml

pietervdvn:
A fresh pull of the master repo, amended with tons of documentation.

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -23,30 +23,361 @@
 
 	<!-- HELP INFORMATION ABOUT FILE -->
 	<!--
-	   Description of blocks:
-	     1) <way attribute="access">
-	        Defines acceptable roads for routingProfile. Value of -1 means way impassable,
-	        value of 1 means way passable.
-	     2) <way attribute="speed" type="speed">
-	        Defines speed limits on roads in km/h (if it is not specified on the road via maxspeed tag)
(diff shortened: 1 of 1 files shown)


vshcherb:
'(Note: actually it doesn't matter which symbol is used. ':height' and '$height' will both call the same parameter. But for clarity, stick to this idiom).'
i don't know where it is coming from but it is simply not true! Look at the code and it will show the difference between '$' and ':'

```
if (value instanceof String && value.toString().startsWith("$")) {
				BitSet mask = tagRuleMask.get(value.toString().substring(1));
				if (mask != null && mask.intersects(types)) {
					BitSet findBit = new BitSet(mask.length());
					findBit.or(mask);
					findBit.and(types);
					int v = findBit.nextSetBit(0);
					o = parseValueFromTag(v, valueType);
				}
			} else if (value instanceof String && value.equals(":incline")) {
				return paramContext.incline;
			} else if (value instanceof String && value.toString().startsWith(":")) {
				String p = ((String) value).substring(1);
				if (paramContext != null && paramContext.vars.containsKey(p)) {
					o = parseValue(paramContext.vars.get(p), valueType);
				}
			}
```

pietervdvn:
Thanks for merging and thanks for pointing out the difference! Finally I know where to start looking on documenting that difference too, I'll do that soonish.

PR #499, opened 2018-07-25
Make routing.xml more consistent: no 'if' or 'gt' as inner elements

pietervdvn:
It bothered me and can be confusing that `if` and `gt` are sometimes used as inner elements of a `select`, so I swapped them all out. This way, routing.xml is more consistent and a bit cleaner.

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -76,27 +76,27 @@
 			<select value="-1" t="highway" v="construction"/>
 			<select value="-1" t="construction" v="yes"/>
 
-			<select value="-1" t="highway" v="motorway">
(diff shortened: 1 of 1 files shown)


vshcherb:
Theoritically I believe it should be no difference though please test on a device to maker sure the calculation time didn't change and there is no weird results. You can also use OsmAndMapCreator but in that case it is good to use with Native library support (only linux)

pietervdvn:
Hey,

It's been four months that I've opened this PR. Is haven't found a chance to follow up on it and I'm probably not gonna have time to test this properly, so I'm closing the PR.

PR #505, opened 2018-08-09
Updated Japanese TTS

meoberto:
After using OsmAnd with the TTS voice in Japanese, I've noticed a few mistakes that I have now corrected in the ttsconfig.p file.

First of all, the prepositions "at", "in", "onto", "toward", etc. in Japanese go at the end of the word they're referring to. Previously the TTS placed them at the beginning of each word, which is ungrammatical.

Then, the 17 lines for each ordinal numbers (1st to 17th) had the ordinal (the 'st' in first, for example) detached from each number. This caused a situation in which English lines were used for the Japanese TTS, which in turn produced situations like "1st." being read as "いちストーンドット"　(1 stone dot). Using the previous detached ordinal, I was not able to get consistent behaviour from my device's TTS. This has been corrected by attaching 番目 (the 'th') to each number.

In turns and roundabouts, the instructions have been made significantly shorter and concise, following the example of most other navigation softwares with Japanese TTS, using e.g. 右方向 (みぎほうこう） to signal a right turn instead of the previous sentences (e.g. 右に曲がって下さい).

There is some initial work done for TTS routing in JavaScript, which is loosely based on the English JS file and incorporates all of the previously mentioned corrections.

### DIFF
--- a/voice/ja/ja_tts.js
+++ b/voice/ja/ja_tts.js
@@ -1,0 +1,755 @@
+
+// IMPLEMENTED (X) or MISSING ( ) FEATURES, (N/A) if not needed in this language:
+//
+// (X) Basic navigation prompts: route (re)calculated (with distance and time support), turns, roundabouts, u-turns, straight/follow, arrival
+// (X) Announce nearby point names (destination / intermediate / GPX waypoint / favorites / POI)
+// (X) Attention prompts: SPEED_CAMERA; SPEED_LIMIT; BORDER_CONTROL; RAILWAY; TRAFFIC_CALMING; TOLL_BOOTH; STOP; PEDESTRIAN; MAXIMUM; TUNNEL
+// (X) Other prompts: gps lost, off route, back to route
+// (X) Street name and prepositions (onto / on / to) and street destination (toward) support
+// (X) Distance unit support (meters / feet / yard)
+// (N/A) Special grammar: (please specify which)
+var metricConst;
+var dictionary = {};
+// TODO set this as a boolena flag to determine if we need to return tts prompts or ogg file names
+var tts;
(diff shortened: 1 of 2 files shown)


voice/ja/ja_tts.js
```diff
@@ -0,0 +1,755 @@
...
+				return (tts ? Math.round(dist/1000.0).toString() : ogg_dist(dist/1000.0)) + " " + dictionary["kilometers"];
+			}
+			break;
+		case "mi-f":
+			if (dist < 160) {
+				return (tts ? Math.round((2*dist/100.0/0.3048)*50).toString() : ogg_dist((2*dist/100.0/0.3048)*5)) + " " + dictionary["feet"];
+			} else if (dist < 241) {
+				return dictionary["1_tenth_of_a_mile"];
+			} else if (dist < 1529) {
+				return tts ? Math.round(dist/161.0).toString() + " " + dictionary["tenths_of_a_mile"] : ogg_dist(dist/161.0) + " tenths_of_a_mile.ogg";
```
PaulStets: Sorry, it was initially my typo. Please change it to ```return (tts ? Math.round(dist/161.0).toString() : ogg_dist((2*dist/100.0/0.3048)*5) + " " + dictionary["tenths_of_a_mile"];

PaulStets:
@meoberto Thank you for contributing. We'll be using the JavaScript TTS in the future. Please update the typo I mentioned in the comment and I'll merge the commit.

meoberto:
The typo has been fixed.

PaulStets:
@meoberto Thanks. I have corrected other minor issues and merged it. The .js file will be also available in the iOS version of the app.

PR #507, opened 2018-08-09
Create sc_tts.js

asereze:
New sc_tts.js file created.

### DIFF
--- a/voice/sc/sc_tts.js
+++ b/voice/sc/sc_tts.js
@@ -7,46 +7,724 @@
 // (X) Other prompts: gps lost, off route, back to route
 // (X) Street name and prepositions (onto / on / to) and street destination (toward) support
 // (X) Distance unit support (meters / feet / yard)
-// (X) Special grammar: distance(nominative/dative), street name (male/female/nothing)
-
-var dictionary = {};
+// (N/A) Special grammar: (please specify which)
 var metricConst;
+var dictionary = {};
 var tts;
 //// STRINGS
 ////////////////////////////////////////////////////////////////
 // ROUTE CALCULATED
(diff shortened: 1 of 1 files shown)


asereze:
@PaulStets I've done it, I hope it's ok now. In case it's not please let me know. I would also like to ask you a thing: is OsmAnd now going to take the numbers (ex "400" for "400 meters") from the .ogg files? Because with the tts voice until now I had the problem that for some reason it pronounced those in Italian instead of Sardinian. Otherwise I would need to find a way to add them as translatable strings like "first" or "seventeenth" are (I guess it would still be the best thing to do, since the voice used for registering the .ogg files and the tts one are quite different).

PaulStets:
I have corrected minor mistakes, so it is ok now. OsmAnd will use the new JavaScript tts logic. The .ogg files are used only when you use the recorded voice prompts. Judging from your other pull request, there were issues with the generated voice files. We will regenerate those files in the future.

asereze:
@PaulStets ok, thank you. Do you already know when the new tts voice files will be available for download in OsmAnd? So that I will be able to check if the numbers will be pronounced correctly or still in Italian.

PaulStets:
@asereze It will be available in the next release. It is already available in the nightly version. By the way, I have tested your file, and it seems to contain non-UTF-8 characters, which causes an error. If you could, please re-encode the file or find non-UTF character an replace it.

asereze:
@PaulStets sorry for the late response, I couldn't use the PC for a while. How can I find the non UTF-8 characters? I've tried opening the file with Notepad++ and it tells me that the encoding is already UTF-8.

PR #518, opened 2018-08-16
add Slovak

Evgenij27:


aceman444:
Hi, is this an automatic conversion of ttsconfig.p from prolog to JS, without logical changes? If so, thank you for doing it. Is this now the primary format and new changes should be done on this file?

PaulStets:
@aceman444 Hi, we have converted the old Prolog files to JavaScript manually trying to preserve the logic. Feel free to edit the file if you notice any mistakes. Starting from the next update we'll be using JS instead of Prolog. You can already test the new engine in the nightly version. We'll be more than glad if you could test it and send us the feedback.

aceman444:
So in OsmAnd 3.1 the Prolog version is still used, but in 3.2 (and nighlties) the JS version is used?

PaulStets:
Yes, you can still switch between the two in the Development settings. When you upgrade to the new version, the js files will be downloaded automatically.

PR #524, opened 2018-08-25
Update ttsconfig.p

aitolos:
remove repeated word


vshcherb:
Not going to be supported

PR #525, opened 2018-08-27
Corrections of Swedish navigation prompts.

goeranu:
This is an extension of #521, which applies the changes previously made to `voice/sv/ttsconfig.p` also to `voice/sv/sv_tts.js`. It also includes a small fine tuning of one message.  I'm not sure if the intention of the reply in #521 was to apply the changes to `sv_tts.js` _also_ or _instead_.  But I figured as long as both files remain in the  repository, it's better if both are more correct.  (I don't dare to do the _git rm_ in a PR; you decide when it is time for that.  After all, the prolog file still worked in the version of the app on my phone.)

The reason for both patches is that the Swedish version of the navigation prompts have some problems. Some of them are simply amusing ("roundabout" is sometimes sounds like "roundabouto" with an "o" in the end), and some are actually misleading (when following one road to get _to_ another, it sounds as you already are _on_ the road you are going to).

With great assistance from Toby Dickenson in the [OsmAnd Google group](https://groups.google.com/forum/#!topic/osmand/0GXtuSublJ0) I've updated the code to try to fix those issues. When testing the prolog version with my installed app, I can't find any more errors.


vshcherb:
voice/sv/ttsconfig.p
 is deleted please correct PR

goeranu:
Updated as requested.

PaulStets:
It's ok, thanks for contributing.

PR #528, opened 2018-09-26
reduce penalty transition

Evgenij27:


aceman444:
Why? Does this tiny change cause any specific change in routing?

maraf24:
Even slight change can affect routing, but it would be good to know what real problem does this solve.

vshcherb:
There should be a test introduced and the link to bug which is pretty obvious.

vshcherb:
Moved to https://github.com/osmandapp/OsmAnd-resources/pull/529/files

PR #542, opened 2019-01-17
Update ttsconfig.p

vooteleaer:
Some translation fixes.


vshcherb:
We don't update prolog files anymore, only js.

PR #546, opened 2019-03-02
Add cobblestone avoidance

pietervdvn:
Hello,

I'm from Bruges. We have a lot of cobblestones, which are not that easy to cycle on. Therefore, I mapped all the surface areas there, and edited the routing profile to have a router.

I hope that adding an extra option in the routing profile will be useful for a lot of people - at least in Bruges and other similar cities it will be useful for others. An due to streetcomplete, the data is becoming complete in other regions as well - so I hope it will be useful around the world. 

Clarification-edit: this PR adds an extra checkbox "avoid sett & cobblestone" in the 'avoid'-section of navigation options. It is disabled by default.

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -854,7 +854,7 @@
 		<parameter id="avoid_stairs" name="Avoid stairs" description="Avoid stairs" type="boolean"/>
 		<parameter id="avoid_borders" name="Avoid border crossing" description="Avoid crossing a border into another country" type="boolean"/>
 		<parameter id="allow_motorway" name="Allow motorways" description="Allow motorways" type="boolean"/>
-		<parameter id="avoid_sett" name="Avoids sett and cobblestone" description="Tries to avoid streets and paths which are paved with sett and cobblestone." type="boolean" default="false"/>
+		<parameter id="avoid_sett" name="Avoid sett and cobblestone" description="Tries to avoid streets and paths which are paved with sett and cobblestone." type="boolean" default="false"/>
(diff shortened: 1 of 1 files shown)


maraf24:
I would use smoothness tag for that.
Also setts can be quite nice surface for cyclist, especially if flattened.

pietervdvn:
In general, they are not that nice to cycle. If they are flattened, there are seperate tags for that - which are more appropriate.

maraf24:
And this separate tag is smoothness.
By the way, we are shifting from using cobblestone to unhewn_cobblestone and your PR doesn't include this value.

pietervdvn:
Please propose a good priority value for the unhewn_cobblestone tag - should it be preferred or discounted?

I do not like smoothness, as it is very subjective. A sett paving can have 'smoothness=good' (as it is quite good to drive over with car), but be quite uncomfortable for a cyclist (needing 'smoothness=intermediate' or even 'bad'). I'm not gonna get started with using this tag - but feel free to add your own routing option.

maraf24:
unhewn_cobblestone is the real cobblestone, so you should include it in your PR.

It is true that sett can be comfortable for car and not for cyclist. This still can be tagged with an  additional tag like bicycle:smoothness.

I was mainly referring to  your explanation for these surfaces as being unfriendly to cyclists, which is not always true.
You can of course make an option to just avoid any sett and cobblestone.

pietervdvn:
I have added 'unhewn_cobblestone' as tag that is avoided.

> You can of course make an option to just avoid any sett and cobblestone.

Which this PR exactly does ;)

pietervdvn:
THANK YOU 😍🎉

I'm looking forward to the new version!

PR #558, opened 2019-05-06
Fix 3-4 hours in sk_tts.js

aceman444:
It seems this typos slipped in at the conversion from prolog. The string is named "hours3_4".

Other than this, the conversion seems OK, and the file is working beautifully in OsmAnd 3.3. Thanks for doing the manual conversion for all languages!

### DIFF
--- a/voice/sk/sk_tts.js
+++ b/voice/sk/sk_tts.js
@@ -281,7 +281,7 @@ function hours(minutes) {
 	} else if (minutes < 180) {
 		return dictionary["hours2"];
 	} else if (minutes < 300) {
-		return Math.floor(minutes/60).toString() + (!tts ? ".ogg " : " ") + dictionary["3_4_hours"];
+		return Math.floor(minutes/60).toString() + (!tts ? ".ogg " : " ") + dictionary["hours3_4"];
 	} else if (tts) {
 		return Math.floor(minutes/60).toString() + " " + dictionary["hours5"];
 	} else if (!tts && minutes < 21 * 60) {


aceman444:
Thank you. Can you also arrange for this file to be updated in the sk voice package (the pre-recorded one, not TTS) ? Even today I can download in OsmAnd one very old one, that does not even have the fix from https://github.com/osmandapp/OsmAnd-resources/commit/ac93683aeeeec526362d7e4afe8a7e311e7b4079 merged, thus is totally busted due to the undefined 'tss' variable.

PaulStets:
@aceman444 Will update the recorded voices tomorrow. Thank you for the fix.

PaulStets:
@aceman444 Please delete the current voice prompts from OsmAnd (Download maps - Local) and install them again. I updated the archive, so everything should work now.

aceman444:
Thanks, downloads properly and works. But can you update the date of the voice package so that all users notice the update via OsmAnd? There is even a Google Play comment mentioning this crash of sk pre-recorded package.

aceman444:
It still does not work. You can see at https://download.osmand.net/list.php that the pre-recorded voices are from 2018 thus users don't get any new files.

The tts shipped with OsmAnd and updated with OsmAnd updates are fine.

PaulStets:
The modification date of the file with recorded prompts wasn't changed deliberately. This doesn't mean that the archive contains the old TTS file. I have checked the archive's TTS file and it is the correct version with all the changes.

aceman444:
OK, but then how are users notified of the change? OsmAnd will not show them there is an update of the file to download.

PR #564, opened 2019-06-14
Improve Arabic TTS

michalfabik:
Special grammar: agreement of number and singular/dual/plural form of unit  
Special grammar: agreement of number and unit gender  
Special grammar: nominative vs. genitive/accusative distinction  
Special grammar: word order with ordinal numbers  
  
Added a lot of strings for numbers to allow for case gender distinction and a function to convert from int's to those strings. Rephrased some instructions to be less verbose. I may have been a little too pedantic with the harakaat in a few places in an attempt to force correct pronunciation (e.g. مئتَيْنِ instead of مئتِينَ).  
Should only affect TTS, a lot of new .ogg's would be needed to achieve the same effect.  
Tested with A Capela and Vocalizer.

### DIFF
--- a/voice/ar/ar_tts.js
+++ b/voice/ar/ar_tts.js
@@ -8,7 +8,7 @@
 // (X) Distance unit support (meters / feet / yard)
 // (X) Special grammar: agreement of number and singular/dual/plural form of unit
 // (X) Special grammar: agreement of number and unit gender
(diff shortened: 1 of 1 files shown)


vshcherb:
Thank you very much, please test in nightly builds and see if it works fine.

PR #565, opened 2019-06-14
Update hu_tts.js

colorimeter:
refactor unneeded long swtiches


vshcherb:
I don't think there is sense in such refactoring cause it is hard to "estimate" statically which dictionary[] should be translated and it is impossible to get it by simple grep function

PR #571, opened 2019-07-08
Update default.render.xml

groschialeux:
Added rendering attributes for track grades 1-5 for route details info.


vshcherb:
Here we need to add strings for translation. Though indeed it should work and now you can create custom rendering types

PR #572, opened 2019-07-12
Translating "bear r/l" as "turn r/l", not "stay on your r/l".

ArenaL5:
The commit message is self-explanatory, Translating “bear” as “keep” is correct for English, but incorrect and confusing in Spanish and Portuguese.

My new clips are copies of `left.ogg` and `right.ogg`.

Is `master` the right branch to point to?

### DIFF
--- a/countries-info/regions.xml
+++ b/countries-info/regions.xml
@@ -134,7 +134,7 @@
 			</region>
 			<region name="hungary" lang="hu" poly_extract="east-europe"/>
 			<region name="moldova" lang="ro" poly_extract="east-europe"/>
-			<region name="poland" inner_download_prefix="$name" lang="pl" poly_extract="east-europe" map="no" roads="yes" join_map_files="yes">
+			<region name="poland" inner_download_prefix="$name" lang="pl" poly_extract="east-europe" map="no" roads="yes" join_map_files="no">
 				<region name="greater-poland" srtm="no" hillshade="no" wiki="no" translate="Greater Poland;entity=node"/>
 				<region name="kuyavian-pomeranian" srtm="no" hillshade="no" wiki="no" translate="Kuyavian-Pomerania;entity=node"/>
 				<region name="lesser-poland" srtm="no" hillshade="no" wiki="no" translate="Lesser Poland;entity=node"/>


vshcherb:
You deleted the files left_bear.ogg but you kept the for recorded voices. So it is quite confusing.
Probably you wnated to use
tts ? "gira a la izquierda" : "**left_keep**.ogg";

ArenaL5:
No, I double checked. I respected the filenames because, in the version you have there are pointers for `left.ogg`, `left_bear.ogg` and `left_keep.ogg`, but the two latter are identical.

The point of this commit is that `left_bear.ogg` and `left_keep.ogg` cannot be identical for Spanish, so I replaced `left_keep.ogg` with a copy of `left.ogg`, to match the text in the JS file. I can delete `left_keep.ogg` and replace it in the JS file with `left.ogg`, if you prefer.

(Also, if you are wondering why are there three participants, I somehow managed to commit to my fork from the wrong account.)

OpenSourceForEver:
I would change "bear" to "turn slightly". I think it is better than "turn"

OpenSourceForEver:
I would also change "on" for "continue". Line 107 says: "// is used if you turn together with your current street, i.e. street name does not change." So I think "continue" is more precise than "on" in spanish.

ArenaL5:
> I would change "bear" to "turn slightly". I think it is better than "turn". Look at http://www.gpspassion.com/forumsen/topic.asp?TOPIC_ID=36357 (message 3).

While that's the meaning of “bear”, I'm not sure OsmAnd doesn't make that distinction... From what I remember, OsmAnd may tell you, as an example `Continue for two kilometres, then keep right` and, then, on the verge of turning, `turn right`/`turn sharply right`/`keep right`, depending on turn angle. `keep right` doesn't imply that the road bends slightly in the first indication.

Please correct me if I'm wrong.

> I would also change "on" for "continue". Line 107 says: "// is used if you turn together with your current street, i.e. street name does not change." So I think "continue" is more precise than "on" in spanish.

This seems sensible.

OpenSourceForEver:
You said in the first message that "bear" was not well translated in Spanish and that it should be "turn" and I am agree with you, the only thing I wanted to clarify is that instead of "turn" I think it is more appropriate " turn sligtly "precisely because of the angle of the turn that you mention yourself, "bear" is supposed to refer to an open angle so I think it should be" turn slightly ".

PR #576, opened 2019-07-26
Improvements in roundabouts

OpenSourceForEver:
To make the instructions shorter when approaching a roundabout I propose to change "in 450 meters enter the roundabout and take the first exit" to "in 450 meters at the roundabout take the first exit". 
In other languages ​​I don't know if it's shorter or not, but at least in Spanish it's shorter and clearer and in English it's slightly shorter


aceman444:
Why the duplicated space in the new string:
" " + " " ?

OpenSourceForEver:
I did not realize. I will fix it. Thanks

PR #577, opened 2019-07-30
Fix mistake last update

OpenSourceForEver:
Fix double  [+ " " +"]

https://github.com/osmandapp/OsmAnd-resources/pull/576#issuecomment-516141105

Thanks to @aceman444

### DIFF
--- a/voice/es/es_tts.js
+++ b/voice/es/es_tts.js
@@ -331,7 +331,7 @@ function roundabout(dist, angle, exit, streetName) {
 	if (dist == -1) {
 		return dictionary["take"] + " " + nth(exit) + " " + dictionary["exit"] + " " + turn_street(streetName);
 	} else {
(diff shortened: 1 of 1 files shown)


OpenSourceForEver:
By the way, what do you think about the change introduced in https://github.com/osmandapp/OsmAnd-resources/pull/576 to do it also in other languages ​​such as English?

aceman444:
I'm not sure. While it is good to have short and concise voice commands, they must be clear to the driver. Each language must make that sure on its own. Not say "enter" (the roundabout) may be ambiguous whether to enter the roundabout or just take a branch right before it.

OpenSourceForEver:
For example "at the roundabout take the second exit" would be clear enought. My englis is not very good but I think "At the roundabout" can not be considered outside the roundabout.

aceman444:
There may be subtle differences even in English, 'at' may not be the right word, 'on' or 'in' could be even better.

PR #578, opened 2019-08-18
Add barrier=sump_buster

gormur:
Currently cars are being routed over this: https://www.openstreetmap.org/node/412369141 That should not be possible.

OSM wiki: https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dsump_buster

This is what they look like around my place: http://stavangerphotobytanty.blogspot.com/2008/07/sporsluse.html


aceman444:
No wonder. OsmAnd will happily route through a ton of barriers (like you have found), because of wrong decision in issue #5170. Adding sump_buster is a temporary solution until you find the next missing barrier.

xmd5a2:
Merged manually

PR #579, opened 2019-08-20
Fix bug: ski routing does not take edges of decorative piste outline anymore

pietervdvn:
When testing the ski routing, I noticed that ski routing would sometimes follow the outline of a ski piste - sometimes even going up onto the mountain...

Faulty behaviour

This pull request adds a tag transform to add an "access=no" tag onto areas, resulting in this correct route planning:

Correct behaviour

Please merge at least one month before the skiing season begins ;p


pietervdvn:
Thanks for merging!

PR #580, opened 2019-08-22
Fix house number showing in LightRS style

RussianFox:


RussianFox:
Sorry. the problem is different

PR #581, opened 2019-08-23
Fix for waterway order in LightRS style

RussianFox:
Default order for waterway is 9, and it more than bridge order and other


xmd5a2:
What issue fixes this pull request? Please provide an example

RussianFox:
I created the issue
https://github.com/osmandapp/Osmand/issues/7489

PR #582, opened 2019-09-02
Fix for waterway and bridge orders in LightRS style

RussianFox:
I made a mistake in the previous pull-request. And now I correct it. Revert order for waterway and add order="10" for bridge area without cycle="true"


RussianFox:
Issue for pull-request https://github.com/osmandapp/Osmand/issues/7489

PR #585, opened 2019-09-12
Removed hard using ogg for 1 hour

RussianFox:
Removed hard using ogg for 1 hour
For this issue https://github.com/osmandapp/Osmand/issues/7546


sonora:
Not a good fix, see my remark at https://github.com/osmandapp/Osmand/issues/7546#issuecomment-531455935

PR #589, opened 2019-10-10
Adjust trunk and trunk_link priority

kylegates:
This fixes routing not using trunk_link roads at the intersections of trunk and primary roads.
The short trunk_link segments were being skipped even though they are shorter and the preferred route of travel.


vshcherb:
There is no plan to do it cause (added comment - https://github.com/osmandapp/OsmAnd-resources/commit/6b17635dcc822176a80d18a86c6ff1af658dfc01). it is done on purpose to avoid unnecessary deviation from straight or curvy motorways, so no links will be used

vshcherb:
We need to fix this issue differently somehow

PR #590, opened 2019-10-11
Bugfix/ski difficulty aerialway

pietervdvn:
Hey @xmd5a2,

It's me again, bothering you with the ski pistes ;p

When testing the new skiing difficulty overview in the nightly, I discovered that the aerialways and connections (few meters of walking between lifts) show up as a quite ugly 'undefined'.

I've slapped on some extra logic in the 'difficulty'-routinfo - but was not able to test this. A few questions that should be checked are:
Pistes

- Do other 'additional'-tags work, apart from 'piste:difficulty'?
- Is having 'tag="aerialway"' correct?

Pistes


pietervdvn:
I have been able to test it now, this is the result:


signal-attachment-2019-10-13-001621

xmd5a2:
aerialway is not a piste difficulty but is closer to piste:type

pietervdvn:
No, that is true.

However, for my use case 'recreational downhill skiing', it is useful to have an overview giving an indication of what the journey will look like (thus: difficulty + where the elevators are). Seen the fact that there is -by definition- no difficulty assigned to aerialways, it would be a logical fit to show 'this is an aerialway here'. I do agree that showing which type of aerialway would be too much.

Also: it is the middle of the night in the weekend. I did not expect an anwser to this before monday morning or even later on!

xmd5a2:
Merged manually https://github.com/osmandapp/OsmAnd-resources/commit/2192e0b66314f698a073a1c241751620d6dc153c

PR #596, opened 2019-11-16
Convert crossing=marked to uncontrolled

aceman444:
iD editor invented its own needless duplicate tag and is polluting the OSM database with it. Let's convert it to the proper value to recognize this type of crossing.


aceman444:
See https://wiki.openstreetmap.org/wiki/Key%3Acrossing#Other_used_tags .

xmd5a2:
Thanks!

aceman444:
Thanks. Will the be also seen as crossing=uncontrolled for the purposes of the routing.xml file?

aceman444:
Also, is any change needed in poi_types.xml, somewhere at
https://github.com/osmandapp/OsmAnd-resources/blob/master/poi/poi_types.xml#L1594 ?

xmd5a2:
crossing=marked is not used in routing.xml. No need to change anything in poi_types.xml also.

aceman444:
Yes it is not used in routing.xml, but as we converted it to 'uncontrolled' in the obf, it should be seen as such by routing.xml.

xmd5a2:
entity_convert works for routing also

aceman444:
Cool, thanks.

PR #598, opened 2019-11-24
routing should consider as well the width tag.

TomSH75:
In Line 500 is properly mentioned maxwidth. Thus in line 503 width has to be mentioned.

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -500,7 +500,7 @@
 			<select value="-1" t="maxwidth"> 
 				<gt value1=":width" value2="$maxwidth" type="length"/> 
 			</select>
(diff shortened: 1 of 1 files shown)


aceman444:
Hi, have you tested it? Or do you just guess from reading the code? Notice that e.g. all 2 blocks have type="length", for all of height and width, which means how to interpret the tag value.
I don't say the patch is wrong I just ask if you checked routing change on a real map file.

TomSH75:
Yes, I have tested it. Actually it is as well an issue of the osmand branch. Please see more details in issue #7978. https://github.com/osmandapp/Osmand/issues/7978

PR #603, opened 2020-01-09
Exit num in nav ui

veliymolfar:


OpenSourceForEver:
Now:
if (dist == -1) {
		return getTurnType(turnType) + " " + dictionary["on"] + " " + dictionary["exit"] + " "
			+ take_exit_name(streetName)
	} else {
		return dictionary["in"] + " " + distance(dist) + " " + getTurnType(turnType)
			+ dictionary["on"] + " " + dictionary["exit"] + " " + take_exit_name(streetName)


Instead, I Suggest the following (at least in Spanish I think it's better):
if (dist == -1) {
		return dictionary["take"] + " " + dictionary["exit"] + " "
			+ take_exit_name(streetName)
	} else {
		return dictionary["in"] + " " + distance(dist) + " " + dictionary["take"] + " " + 
                dictionary["exit"] + " " + take_exit_name(streetName)

vshcherb:
Yes we're going to redo it to avoid some wrong translations and leave the old version in case we don't know

PR #604, opened 2020-01-12
Improvements on "exit num", change "bear" and delete unnecessary lines

OpenSourceForEver:
["left_bear"] and ["right_bear"] are better translated as "turn left" and "turn right" as said in https://github.com/osmandapp/OsmAnd-resources/pull/572#issuecomment-515243049.

suggestions for "exit num" are more clear and shorter.


vshcherb:
We're going to update voice instructions about exit turns this week

PR #625, opened 2020-06-19
added traffic signal/signs direction tags

MadWasp79:


vshcherb:
Wrong tags

aceman444:
On highway=give_way and highway=stop the tag is only "direction", see https://wiki.openstreetmap.org/wiki/Tag:highway%3Dstop

PR #631, opened 2020-07-13
Correct German words: Strasse and Fusgänger

jaller94:
* Strasse is ok in Swiss German and in Liechtenstein, but nowhere else. [Straße](https://de.wiktionary.org/wiki/Stra%C3%9Fe) is the more common spelling.
* Fusgänger is a misspelling of [Fußgänger](https://de.wikipedia.org/wiki/Fu%C3%9Fg%C3%A4nger%C3%BCbergang).

### DIFF
--- a/voice/de/de_tts.js
+++ b/voice/de/de_tts.js
@@ -70,7 +70,7 @@ function populateDictionary(tts) {
 	
 	// STRAIGHT/FOLLOW
 	dictionary["go_ahead"] = tts ? "Weiter geradeaus" : "go_ahead.ogg";
-	dictionary["follow1"] = tts ? "Dem Strassenverlauf" : "follow1.ogg";
+	dictionary["follow1"] = tts ? "Dem Straßenverlauf" : "follow1.ogg";
 	dictionary["follow2"] = tts ? "folgen" : "follow2.ogg";
 	
 	// ARRIVE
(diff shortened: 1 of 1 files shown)


vshcherb:
cc @sonora 
I think there was a purpose to support TTS-Engines that don't work with that letter, though I'm not sure what was exact reasoning.

sonora:
The question is do you HEAR any wrong pronunciation with how things are solved today? (IWith all TTS engines and versions I have things are perfectly ok.)

The file we are dealing with here governs the  pronunciation for the  TTS engine, meaning if something is correct spelling or not is not the primary focus.

While a good TTS engine should normally know how to pronounce a word when correctly spelled in that language, apparently at the time I implemented that solution, so many TTS engines had severe intonation issues that that workaround was necessary.

Things may have improved with better TTS engines today, but Ipropose there is not sufficient reason to change the current solution (and re-do lots of testing on many devices) if it works flawlessly?

sonora:
PS: I just checked, also elsewhere in our code I do a general replacement of all 'ß' by 'ss' before passing on to the TTS engine to aid pronouncing German street names (many contain 'Straße'), because there are engines out there not handling a 'ß' well.

And the issue with using Fussgänger' is that some engines then pronounce a very short 'u," more like in the word 'Bus', and not like in the word 'Bußgeld', as it shoul be.

PR #632, opened 2020-07-23
FIXING: osmc:symbol processing happens after tag transformation, that…

ubovics:
… overwrites transformed hiking symbols


vshcherb:
osmc:symbol will be split before tag transformation, so this is the right order https://github.com/osmandapp/OsmAnd-resources/commit/2cf72018ca10d8b4b94046d9eceef00a4097f4d8

PR #637, opened 2020-08-23
maxDefaultSpeed for public transport

OpenSourceForEver:
bus speed= "$maxspeed" requires maxDefaultSpeed to be 95.

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -2094,8 +2094,8 @@
 		</point>
 	</routingProfile>
 
-	<!-- max default speed is max speed for any transort, min speed is pedestrian speed -->
-	<routingProfile name="public_transport" minDefaultSpeed="4.0" maxDefaultSpeed="80.0" baseProfile="public_transport">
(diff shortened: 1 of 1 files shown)


aceman444:
Why? What does this fix? Which is the issue #637 ? There does not seem to be such a report.

Depends on what public transport vehicle you mean. Maxspeed 95 for a bus/tram/trolleybus seems too high, for a train it could be much higher. But the average speed may be lower, due to stopping at stations/ bus stops.

OpenSourceForEver:
This only changes "maxDefaultSpeed" to 95. It only change "maxspeed" to buses because bus speed="$maxspeed".

Other transport types has their own speed limit so this only changes speed limit buses. Now speed limit buses is "$maxspeed" but not higher than 95 km/h

aceman444:
Ok, I see, thanks, https://github.com/osmandapp/OsmAnd-resources/blob/4944f63544ab6a4b3ceb44d9dcf952d5bd64fb67/routing/routing.xml#L2180 .

PR #642, opened 2020-09-14
ferry fixed for bicycle routing

MadWasp79:


aceman444:
What is this fixing? Taking a ferry always takes a large penalty, the ferry may not be available immediately.

vshcherb:
Speed of ferry shouldn't be lower of actual ferry speed

PR #646, opened 2020-10-04
Realistic and brief version

Freedoo:
If your wife told you "Den Kreisverkehr an der dritten Ausfahrt verlassen" you'd probably ask her if she feels OK... No German speaker would use the phrases generated by the current de_tts, at best a lawyer or a policeman would do.

A little bid annoyed by the talkative Osmand I made much of the commands shorter and nearer to real life: (i.E. "Die Entfernung beträgt 42 km, Zeit 25 Minuten" --> "42 km in 25 MInuten" , "Den Kreisverkehr an der zweiten Ausfahrt verlassen" --> "Im Kreisel zweite raus.", "In 400m rechts abbiegen" --> "In 400m rechts", "Sie überschreiten die Höchstgeschwindigkeit" --> "Zu schnell", "zirka einen Kilometer" --> "rund einen Kilometer", "Geschwindigkeitskontrolle" --> "Blitzer". ...)

Try it - you won't miss anything!

### DIFF
--- a/de_tts.js
+++ b/de_tts.js
@@ -1,0 +1,698 @@
+//  More realistic, brief and natural commands										 
+
+var dictionary = {};
+var metricConst;
+var tts;
+
+//// STRINGS
+////////////////////////////////////////////////////////////////
+// ROUTE CALCULATED
+function populateDictionary(tts) {
+
+	dictionary["route_is1"] = tts ? "Route berechnet. " : "route_is1.ogg";
+	dictionary["route_is2"] = tts ? " " : "route_is2.ogg";
+	dictionary["route_calculate"] = tts ? "Route neu berechnet. " : "route_calculate.ogg";
+//	dictionary["distance"] = tts ? ", Entfernung " : "distance.ogg";
+dictionary["distance"] = tts ? ", " : "distance.ogg";
+
+
+
+	// LEFT/RIGHT
+	dictionary["prepare"] = tts ? "Bald   " : "prepare.ogg";  // bald sounds better then Vorbereiten zum
+	dictionary["after"] = tts ? "in " : "after.ogg";
+	
+	dictionary["left"] = tts ? "links" : "left.ogg";
+	dictionary["left_sh"] = tts ? "scharf links" : "left_sh.ogg";
+	dictionary["left_sl"] = tts ? "leicht links" : "left_sl.ogg";
+	dictionary["right"] = tts ? "rechts" : "right.ogg";
+	dictionary["right_sh"] = tts ? "scharf rechts" : "right_sh.ogg";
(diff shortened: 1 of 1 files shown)


Lukas458:
Actually, I don't know what should be wrong with "Den Kreisverkehr an der dritten Ausfahrt verlassen". It's a clear annoucement for me. I think we do net need to take over every "shortspeech". But people might have different opinions on that.

Freedoo:
There's nothing wrong, but it is as well unnecessary long as unusal / uncommon. 
I never doubted tha it's a clear annoucement, but in reality nobody (except a policeman or a lawyer) would say so. In my opinion a navigation system should not talk more then necessary and should not speak like a policeman. This is in particular the case when you have a conversation in your car or listening to music.

vshcherb:
I think the target folder is wrong, anyway we could keep it under name de-brief_tts.js next to original de_tts.js

Freedoo:
I think, that's a good idea, so one can choose.
Could you perhaps move it to the right folder as I am not so familiar with github, sorry?

vshcherb:
I will let @sonora review a german version. The correct folder will be here https://github.com/osmandapp/OsmAnd-resources/tree/master/voice/de

sonora:
The old Nokia navigation system used to have a voice called 'Surfer dude', which did not only sport a California accent, but also some cool wording.

I will review this and implement as a second German option, either as de_brief as Victor suggested, or maybe as de_casual or whatever seems appropiate.

sonora:
Thanks again for the PR! Please see the above referenced commit, I think a test version should now become available in the downloads.

My comments, also in the PR:

- Unfortunately that PR was based on a very outdated de-tts file, which still contained many issues (like too much chattiness, an issue with rounding (affecting the associated recorded voice), no support yet for highway exit prompts, etc.)

- I have ported the ideas to the latest tts standard and fixed some issues (needed to separate "roundabout_exit" from "exit", etc.)

- After testing, I really quite like the ideas implemented. In practice, however, I also realize that too curt voice prompts (e.g. one word only turn instructions like "left") may be harder to understand (especially with background noise or when you are in conversation) than if you have a short sentence. I have therefore for now committed this as an alternative called "de-casual".

- Please test! If there is majority feedback finding this really better than the current de_tts default, we can think of swapping or even replacing, ok?

- It is worth noting that the turn pre-announcements are blanked out, "Test voice prompts" buttons 3.1, 3.4, 4.1 and 5.1, just in case anybody testing is wondering.

aceman444:
Also notice there already exists a hu-formal language at https://download.osmand.net/list.php so Hungarian already has 2 variants, similar to what you are trying to do here for German.

sonora:
Yes, probably similar. We can leave them both, or combine if we find commonly accepted middle ground, let's see.

PR #647, opened 2020-10-07
Update regions.xml

damjang:
Change the translate of Italy regions to get translation from relations (because there are more translations in relations as in nodes). 4 regions that are more sensitive to name change because of bilingual/trilingual names are translated from en name, that is more stable.


xmd5a2:
Have you tested it?

damjang:
Sorry, no. How can I do it?

xmd5a2:
It can be done if you build regions.ocbf
OsmAndMapCreator-main/utilities.sh generate-ocbf /dir/  > /dir/regions.log
and checking for "!!!" string in regions.log
But I already did that

damjang:
Thank you!

PR #655, opened 2020-12-03
Grammer and pronounciation corrections

madanadam:
Corrected grammer errors and  order at sentences

### DIFF
--- a/voice/tr/tr_tts.js
+++ b/voice/tr/tr_tts.js
@@ -18,7 +18,7 @@ var tts;
 function populateDictionary(tts) {
 	// ROUTE CALCULATED
 	dictionary["route_is"] = tts ? tts ? "Yolculuk" : "route_is.ogg" : "route_is.ogg";
-	dictionary["route_calculate"] = tts ? "Rota yeniden hesaplandı" : "route_calculate.ogg";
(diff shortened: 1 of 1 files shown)


sonora:
Hi, thanks for working on this!

It looks like this is a massive change, both to wording and grammar. Was the previous so bad that this should replace it? Or is this maybe some regional variant or alternative we should somehow name and have in addition?

madanadam:
It was weird to listen in Turkish especially road names. So I want to change order of words. I wonder who did previous translation? Because it was incorrect in Turkish. I live in Turkey and my main language is Turkish. After this corrections I will change other files Turkish files

sonora:
Ok, thanks!

sonora:
PS: You have added many leading or trailing blanks in the dictionary section, I believe these are all obsolete. I had explicitly added needed separators to all TTS' grammar sections many months ago.

madanadam:
Its because of my editor i think :)

sonora:
I've cleaned it up for you. If there is anything not as expected, please let me know here, ok?

madanadam:
Thanks a lot. I'll pay more attention to spaces in my other works.

PR #658, opened 2020-12-28
#10230 Add sharpTurn and slightTurn, change penalty for sharpTurn

alisa911:


routing/routing.xml
```diff
@@ -15,7 +15,7 @@
 
 	<!-- 1.2 Build A* graph in backward/forward direction (can affect results) -->
 	<!-- 0 - 2 ways, 1 - direct way, -1 - reverse way -->
-	<attribute name="planRoadDirection" value="0" />
+	<attribute name="planRoadDirection" value="1" />
```
vshcherb, line 18: wrong change

routing/routing.xml
```diff
@@ -902,7 +902,7 @@
 		</point>
 	</routingProfile>
 
-	<routingProfile name="bicycle" baseProfile="bicycle" restrictionsAware="true" minSpeed="5" defaultSpeed="10" maxSpeed="33" leftTurn="6" rightTurn="6" followSpeedLimitations="false" onewayAware="true" heuristicCoefficient="1.4">
+	<routingProfile name="bicycle" baseProfile="bicycle" restrictionsAware="true" minSpeed="5" defaultSpeed="10" maxSpeed="33" sharpTurn="6" easyTurn="6" followSpeedLimitations="false" onewayAware="true" heuristicCoefficient="1.4">
```
vshcherb, line 905: easyTurn or slightTurn? 
p.s. should be slightTurn

aceman444:
HI, what is this change doing? Why are right and left turns removed? Turning right is often cheaper in right-side driving countries than left turn (thus crossing the opposite direction lane). Why is this removed?

vshcherb:
they are not removed and only renamed.
Since 2015 leftTurn meant sharp turn (not left or right)! It was already in 2015 there was an issue to detect leftTurn in left-driving countries. So it was replaced with same value and different meaning.

So only now wording finally reflects reality

aceman444:
OK, thanks, even though real rightTurn might still be useful.

PR #670, opened 2021-02-10
Added new fields in OBF.proto: preciseDx, preciseDy for POI

ivanPyrohivskyi:
Added new fields in OBF.proto: preciseDx, preciseDy for POI

### DIFF
--- a/protos/OBF.proto
+++ b/protos/OBF.proto
@@ -609,8 +609,7 @@ message OsmAndPoiBoxDataAtom {
 
   repeated uint32 textCategories = 14; // v1.7
   repeated string textValues = 15;
-  optional sint32 preciseDx = 16; // delta encoded to OsmAndPoiBox on 26 zoom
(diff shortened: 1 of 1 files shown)


vshcherb:
More comments is needed

protos/OBF.proto
```diff
@@ -609,6 +609,7 @@ message OsmAndPoiBoxDataAtom {
 
   repeated uint32 textCategories = 14; // v1.7
   repeated string textValues = 15;
+  optional int32 precisionXY = 16; // precision in 1-xy-xy-xy binary format
```
vshcherb, line 612: This field is variable-length precision-improvement from 24th zoom up to X zoom. It's encoded by interleaving next zoom level bits with a high bit prefix of "1". So, precisionXY from 24 -> 25 zoom could be 3 any bit number: 110, 111, 100,  101. From 24 -> 26 - 5 bit number, etc...

PR #674, opened 2021-03-05
Add low_emission_zone tag

Dima-1:


aceman444:
Hi, how is this supposed to work? There are no low_emission_zone=yes tags on ways so the addition to routing.xml seems without any effect. Low emission zones are mapped as boundaries encompassing the roads, thus OsmAnd would need to calculate which roads belong into the low emission zone, see https://wiki.openstreetmap.org/wiki/Tag:boundary%3Dlow_emission_zone .

PR #675, opened 2021-04-01
Penalty for traffic lights

OpenSourceForEver:
Changing delay for traffic signals to 10/15 seconds OsmAnd calculates more accurate times.

Related https://github.com/osmandapp/OsmAnd/issues/6539#issuecomment-810153187

I have calculated some city routes and compared with real time (driving with my car without traffic jams) and with other navigation softwares (real time and other navigation softwares are similar in time. Two examples:

From https://www.openstreetmap.org/#map=19/40.42446/-3.71176 to https://www.openstreetmap.org/#map=19/40.43502/-3.71932    1.3 km
Other softwares: 3 min
OsmAnd (30 seconds per traffic light): 10 min
OsmAnd (10 seconds per traffic light): 4 min

From https://www.openstreetmap.org/#map=19/40.41940/-3.69275 to https://www.openstreetmap.org/directions?from=40.41937%2C-3.69276&to=#map=18/40.47888/-3.68574   6,8 km
Other softwares: 12 min
OsmAnd (30 seconds per traffic light): 23 min
OsmAnd (10 seconds per traffic light): 17 min

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -777,7 +777,7 @@
 			<select value="25" t="barrier" v="toll_booth"/>
 			<select value="25" t="barrier"/>
 			<select value="10" t="traffic_calming"/>
-			<select value="30" t="highway" v="traffic_signals"/>
+			<select value="10" t="highway" v="traffic_signals"/>
 			<select value="1"  t="crossing" v="unmarked"/>
 			<select value="5"  t="crossing" v="uncontrolled"/>
 			<select value="15" t="highway" v="crossing"/>


aceman444:
This is surely city-dependent, in some cities there may be red signal on a traffic light for 60 seconds. Of course, OsmAnd uses some average value, maybe that's why it has 30 seconds. So changing to 10 seconds is just a random tweak that will regress in cities others that you have tested.

More useful would be if OsmAnd respected the traffic_signals:direction tags and used other heuristics to ignore signals and stop/give_way signs that do not apply in the current driving directions. I think there are issues filed for those.

xmd5a2:
> More useful would be if OsmAnd respected the traffic_signals:direction tags
This is already supported

OpenSourceForEver:
> This is surely city-dependent, in some cities there may be red signal on a traffic light for 60 seconds.

In my city traffic lights are red for 60 seconds, but it does not mean that I have to wait 60 seconds at all traffic lights, not all the traffic lights are red when I go through them and not all the red lights I have to wait 60 seconds. 

Therefore, taking into account not only my calculations, but also https://github.com/osmandapp/OsmAnd/issues/6539#issuecomment-810153187 and https://github.com/osmandapp/OsmAnd/issues/10251
https://github.com/osmandapp/OsmAnd/issues/2625
I think it would be better to reduce the delay for traffic lights. If you think that 10 seconds is a drastic reduction, you could set 15 or 20 seconds but 10 seconds would give a better ETA.

vshcherb:
I'm afraid it sounds quite a radical change, mostly it is 20-30 seconds but it varies on a city. I think if we try to solve group of traffic signals issue (dont' count same penalty twice) that would be a wrong approach. I personally think it should not be below 20-25 seconds

Zero3:
Just adding some more anecdata: From my personal experience so far, 30 seconds is also too much for Denmark where I live. Here, traffic lights are quite optimized, and many also have sensors that make the lights go green on approach. I am working on a Denmark-specific routing.xml configuration, and I currently use 15 seconds for it, which seems to be a more realistic value.

The point about this varying from place to place is valid. Perhaps it should be possible to configure this as a user somewhere in the GUI. Or alternatively, make it possible to provide country-specific defaults (there are a lot of other values in routing.xml that make no sense for Denmark either, for example).

vshcherb:
Unfortunately we don't have regional configuration yet in routing.xml, it sounds reasonable to have it though

aceman444:
@vshcherb Yes, that would be really great as it would solve many problems, like different max. speeds per country, access defaults, etc.

Zero3:
Also see this comment in another issue about wrong time estimation in Denmark, where I did the math on traffic light timing: https://github.com/osmandapp/OsmAnd/issues/10251#issuecomment-926008252. Based on this, and other traces I analyzed, I believe a reduction to **20 seconds** would be appropriate for Denmark.

danepowell:
I understand this is a regional variation to some extent so maybe a bunch of "me toos" are not helpful... but fwiw, I agree with this change, 30s is much too high in most cities under normal driving conditions.

I'm guessing that traffic light wait time is inversely correlated with traffic light density along a route, as well as with the size of the road being traveled. In a dense city, lights are typically synchronized and have a duty cycle that minimizes wait times along larger roads. Maybe OsmAnd could use this information to dynamically predict wait times rather than using a fixed value.

vshcherb:
Following many discussions we can reduce it to 20 seconds but anyway it's better to have some scientific approach rather than experimental cause we can't have enough experiments I believe.

OpenSourceForEver:
Although it is true that it depends on the city or the country, there seems to be a consensus that 30 seconds is an excessive amount, reducing this figure, even slightly, seems to give better results for everyone.

A "scientific" reason for this is the fact that not all traffic lights are red in our path and of all those that are red, we do not have to wait 100% of the entire time that the traffic light is red because it's in red when we get to it.

vshcherb:
1st route - https://osmand.net/map?start=40.424460%2C-3.711760&end=40.435020%2C-3.719320&mode=car#15/40.4275/-3.7110 - 1.4 km 9min - 15 traffic signals
2nd route - https://osmand.net/map?start=40.419400%2C-3.692750&end=40.478880%2C-3.685740&mode=car#13/40.4487/-3.6597 - 6.9 km 23 min

vshcherb:
The main issue here that mapping is not self-sufficient, this traffic signal shouldn't be part of the route https://www.openstreetmap.org/node/1179022561. It's actually the same as https://www.openstreetmap.org/node/1179022127.

As I understand from documentation we could ignore combination of crossing = traffic_signals
> If the traffic signal is not separately mapped, the node that is the actual crossing between the road and [highway](https://wiki.openstreetmap.org/wiki/Key:highway)=footway/pedestrian should be tagged with highway=[crossing](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dcrossing) + [crossing](https://wiki.openstreetmap.org/wiki/Key:crossing)=[traffic_signals](https://wiki.openstreetmap.org/wiki/Tag:crossing%3Dtraffic_signals).

ivanPyrohivskyi:
20

xmd5a2:
20 seconds for traffic light gives much more closer to Google routing times

xmd5a2:
Merged manually

PR #682, opened 2021-04-26
fix test 393

alisa911:


test-resources/search/depo.json
```diff
@@ -0,0 +1,20 @@
...
+        "totalLimit": -1,
+        "lang": "",
+        "transliterateIfMissing": false,
+        "emptyQueryAllowed": false,
+        "sortByName": false
+    },
+    "phrase": "Depo",
+    "results": [
+        "Depot (Landuse) [[1, POI_TYPE, 1.000, 0.00 km]]",
+        "Depo [[1, POI, 1.000, 0.04 km]]",
```
vshcherb, line 15: Change test so "Depo" is further than "Depot" but it still should be higher in the list

PR #685, opened 2021-05-06
Add transportRoute tags

ivanPyrohivskyi:


vshcherb:
Idea:
/// Tag / value array for transport route (names, interval, …)
// encoded reference to tag / values from transport stringtable
repeated uint32 attributeTagValueIds = 19;
// bytes array = varint32 (reference of tag in string table) + UTF-8 name bytes
repeated bytes attributeTextTagValues = 20;

protos/OBF.proto
```diff
@@ -357,6 +357,8 @@ message TransportRoute {
 	optional bytes geometry = 17;
 
 	repeated TransportRouteSchedule scheduleTrip = 18;
+	repeated uint32 attributeTagIds = 19;
```
vshcherb, line 360: 0 comments ???

protos/OBF.proto
```diff
@@ -357,6 +357,8 @@ message TransportRoute {
...
+	repeated bytes attributeTextTagValues = 20;
```
vshcherb, line 361: 0 comments ???

PR #691, opened 2021-06-01
Adding TTS for Indonesian

ajyto:
I'm adding TTS for Indonesian Language

### DIFF
--- a/voice/id/id_tts.js
+++ b/voice/id/id_tts.js
@@ -17,119 +17,119 @@ var tts;
 ////////////////////////////////////////////////////////////////
 function populateDictionary(tts) {
 	// ROUTE CALCULATED
-	dictionary["route_is"] = tts ? "The trip is" : "route_is.ogg";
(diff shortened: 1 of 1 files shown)


vshcherb:
Is this file tested with Development Settings -> Test Voice Prompts, Navigation? Do you like the result ? And which TTS-engine did you use? 
It might need to be added https://github.com/osmandapp/OsmAnd/blob/master/OsmAnd/assets/bundled_assets.xml#L33 later

ajyto (commented):
Adding commas, moving "nth" and "exit" into more logical position on indonesian

ajyto:
> Is this file tested with Development Settings -> Test Voice Prompts, Navigation? Do you like the result ? And which TTS-engine did you use?

after a few tweaks, yes. today i commit new revision that flows better. i use ordinary google / android TTS that comes with Android 11

> It might need to be added https://github.com/osmandapp/OsmAnd/blob/master/OsmAnd/assets/bundled_assets.xml#L33 later

i haven't contribute to this file yet. i think the bundling can come after the tts considered finished.

PR #693, opened 2021-06-14
Fix obf tracks selectable

Dima-1:


poi/poi_types.xml
```diff
@@ -4492,8 +4492,9 @@
 				<poi_additional name="max_speed" tag="max_speed" type="text"/>
 				<poi_additional name="min_speed" tag="min_speed" type="text"/>
 				<poi_additional name="route_activity_type" tag="route_activity_type" type="text"/>
+				<poi_additional name="route_radius" tag="route_radius" type="text"/>
```
vshcherb, line 4495: should be additional

vshcherb: we should have limited number of values: 5, 10, 50, 100, 500, 1000, 5000, 10000

PR #697, opened 2021-07-29
Fix and test long deviation for motorway exit

Dima-1:


Dima-1:
osmandapp/OsmAnd#12066

PR #707, opened 2021-09-02
enable motorcycle=* as additional rendering filter

cmoffroad:


cmoffroad:
Thanks! How long should I expect until the changes are applied in a country dataset via OSM live and original download?

PR #709, opened 2021-09-03
Support WikiVoyage listing type Go

Dima-1:


vshcherb:
where is color / icon support?

PR #710, opened 2021-09-06
Add conveying tag in pedestrian routing profile

n-timofeev:
https://wiki.openstreetmap.org/wiki/Key:conveying


vshcherb:
We need to also check if this tag is present in rendering_types.xml

PR #711, opened 2021-09-06
enable survey=ground with alternative source:position

cmoffroad:


vshcherb:
There is no description could enlight what's the issue (link) or discussion are you trying to solve with this request?

cmoffroad:
Sorry Victor, communication regarding this topic was kept in the public Telegram chat:

> It would be very valuable to know whether a road geometry is based on ground survey vs imagery. 
> I am building a rendering style that will be used mostly for scouting unpaved roads in areas where there has been little ground survey done. 
> If I could render differently roads that have been surveyed on the ground (accurate) vs imagery (may be inaccurate), it would be perfect. Similar to the GPS traces overlay in ID.

By testing `source:geometry` and here `source:position` (also commonly used) against values like `GPS`, `survey` and `mapillary`, we can attach `survey=ground` to matching ways during OBF creation process, and this can be then used as  `additional` filter condition by rendering styles implementation.

I can confirm the first `source:geometry` integration works great and my plugin renderer can effectively adapt styles based on this implementation.

PR #713, opened 2021-09-07
New: Horsebackriding

derfuessl:
Add Routing for Horsebackriding

OSMand would be the ideal navigation system for horse riding as all the necessary data is contained in OSM.
There are currently no comparable navigation systems on the market.

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -1884,6 +1884,292 @@
 		</way>
 	</routingProfile>
 
+	
+		
+		<routingProfile name="horsebackriding" baseProfile="horsebackriding" restrictionsAware="true" minSpeed="2" defaultSpeed="6" maxSpeed="30" sharpTurn="0" slightTurn="0" followSpeedLimitations="false" onewayAware="false">
+
+		<attribute name="heuristicCoefficient" value="1.2" />
+		<!-- NEW ROUTING API -->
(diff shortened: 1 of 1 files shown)


vshcherb:
Looks cool and interesting

vshcherb:
Did you participate any Telegram chat / discussing specific about development?

derfuessl:
No. Not yet. I just found the import routing.xml option in the OSMand-App last week an started directly to make one for riding. Since I have a bit experience in programming and used OSM a lot, making the xml was not so hard.I was looking for something like this for some time. I have tried the xml File and was very happy whit the results. (One additional option for the rendering would be nice, but this is for another day ;-) ) After looking for a suitable program for a long time, I wanted to share my results.And make the option of an (German) Translation for the Parameter in Profile.
-------- Ursprüngliche Nachricht --------Von: vshcherb ***@***.***> Datum: 07.09.21  21:09  (GMT+01:00) An: osmandapp/OsmAnd-resources ***@***.***> Cc: derfuessl ***@***.***>, Author ***@***.***> Betreff: Re: [osmandapp/OsmAnd-resources] New: Horsebackriding (#713) 
Did you participate any Telegram chat / discussing specific about development?

vshcherb:
Sounds cool, CC @sonora @xmd5a2

PR #732, opened 2021-10-13
Make the track dash style patterns more consistent with OSM Mapnik

martenjj:
For the styles "default" (from which most others are derived), "topo"
and "touring".

Addresses https://github.com/osmandapp/OsmAnd/issues/12953

Rendering of same location as in #12953 with updated rendering files installed:

osmand-new-Screenshot_20211007-194909_OsmAnd+


vshcherb:
Can't accept it. It touches 3 rendering styles of different authors (@sonora - touring, @xmd5a2 - default / topo).

We can continue discussion here and it could be merged partially. Though goal of unification is very unclear cause dash patterns could be less or more visible within certain styles

PR #733, opened 2021-10-15
Issue 12911

ivanPyrohivskyi:
Minimal needed changes for fix https://github.com/osmandapp/OsmAnd/issues/12911


vshcherb:
It's better to increase priority "bicycle=yes" cause "highway" = "pedestrian" doesn't assume bicycle are allowed or different than pedestrian, so speed should not be bigger than 7 kmh (pedestrian speed)

PR #735, opened 2021-10-20
Set no access for cycleway for car routing

ivanPyrohivskyi:
Set no access for cycleway for car routing before access=yes
https://github.com/osmandapp/OsmAnd/issues/13020


aceman444:
What if there is access=yes or motor_vehicle=yes on the cycleway? There may be segments of cycleways where vehicles may have allowed access. The rule disallowing cycleways for car should be placed in the file after the 'access' rules are evaluated.

PR #750, opened 2021-11-10
New tags, rendering and routing changes

vlzuykov:
1. Adding to obf tag "cutting" for rendering as well as some tags which can be useful for routing
2. Add render style for "cutting" tag (the same style as for "embankment"="yes")
3. Add new tag for tram rails and highway crossing ("tram_level_crossing") as an obstacle for bikes

### DIFF
--- a/obf_creation/rendering_types.xml
+++ b/obf_creation/rendering_types.xml
@@ -3107,6 +3107,7 @@
 		<type tag="railway" value="funicular" minzoom="13" />
 		<type tag="railway" value="halt" minzoom="12" order="10"/>
 		<type tag="railway" value="level_crossing" minzoom="15" />
+        <type tag="railway" value="tram_level_crossing" minzoom="15" />
 		<type tag="railway" value="light_rail" minzoom="11"/>
 		<type tag="railway" value="monorail" minzoom="13" />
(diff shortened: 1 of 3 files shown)


obf_creation/rendering_types.xml
```diff
@@ -3107,6 +3107,7 @@
 		<type tag="railway" value="funicular" minzoom="13" />
 		<type tag="railway" value="halt" minzoom="12" order="10"/>
 		<type tag="railway" value="level_crossing" minzoom="15" />
```
vshcherb, line 3109: Formatting: something wrong with tabing or spacing.

rendering_styles/default.render.xml
```diff
@@ -11770,6 +11770,7 @@
 			<case tag="man_made" value="dyke"/>
 			<case engine_v1="true" tag="man_made" value="embankment"/>
 			<case tag="embankment" value="yes"/>
+            <case tag="cutting" value="yes"/>
```
xmd5a2, line 11773: This is not correct. Our rendering engines can't correctly render cutting at that moment.

xmd5a2:
Merged manually

PR #751, opened 2021-11-19
Add renderingPriority property

ivanPyrohivskyi:


rendering_styles/default.render.xml
```diff
@@ -3411,91 +3411,91 @@
...
-					<case tag="highway" value="road" order="79">
+					<case tag="highway" value="road" order="79" renderingPriority="5">
 						<apply_if layer="1" order="102"/>
 						<apply_if additional="layer=2" order="132"/>
 						<apply_if additional="layer=3" order="160"/>
 						<apply_if layer="-1" order="57"/>
 						<apply_if additional="layer=-2" order="36"/>
 					</case>
-					<case tag="highway" value="unclassified" order="79">
+					<case tag="highway" value="unclassified" order="79" renderingPriority="5">
```
vshcherb, line 3491: default value not needed to be specified and better to make default value - 50 to have more flexibity cause this number is int.

aceman444:
Hi, so what is this feature about? There is no description.

vshcherb:
it was reverted but indeed we need to add comments

PR #752, opened 2021-11-26
Rollback to old rendering

ivanPyrohivskyi:
Remove densityPriority. Add special rule for tunnel


vshcherb:
Not needed as not supported param

PR #759, opened 2021-12-22
motor_vehicle:backward = no

ivanPyrohivskyi:
Add oneway rule motor_vehicle:backward = no, motor_vehicle:forward = destination for car routing


routing/routing.xml
```diff
@@ -551,6 +551,7 @@
 			<select value="-1" t="oneway" v="-1"/>
 			<select value="1" t="roundabout"/>
 			<select value="1" t="junction" v="roundabout"/>
+			<select value="1" t="motor_vehicle:backward" v="no"/>
```
vshcherb, line 554: - Missing ```motor_vehicle:forward``` ?

aceman444:
What about vehicle:backward/forward ? This could also go into the bicycle routing rules.

PR #766, opened 2022-01-04
Update de_tts.js: Don’t read out ref= tags for lower-tier highways, if a street name is available.

luziferius:
Do not read out aloud ref=* values for highway=secondary and highway=tertiary, if a street name is available.

# Background and reasoning

Major roads highway=(motorway|primary|secondary|tertiary) in Germany 
have ref tags containing a type letter (A|B|L|K) respectively and number, 
e.g. "ref=K 673" for a tertiary road. But only refs for motorways and highway=primary are
publicly visible on signposts, with motorways being blue and primary roads being yellow:
refs

The road numbers for secondary and tertiary highways are typically not printed on signs, especially inside cities.
It _may_ appear in rural areas if the street explicitly has no name, but I can’t remember having seen one ever.

So reading the ref value of secondary or tertiary roads out loud
does not benefit the driver, as the information can not be cross-referenced with the
visible signs. It increases the verbiage for no benefit.

The proposed change causes the TTS to omit reading out the ref tag for lower-grade roads, if a street name is available.

As a fallback, it still keeps the ref as the name, if no street name is available, because _any_ name is better than _none_.

As a test case, here is a [node of a crossing](https://www.openstreetmap.org/node/2297893627) that touches a primary highway, a tertiary highway with a name and a tertiary highway without name.

Turning east/west reads out “turn right on B1 Werler Straße”. (Reads out both ref and name)
Turning north reads out “turn left on Hemmerder Dorfstraße”. (Only reads out the name)
Turning south reads out “turn left on K35” (Uses the ref as a fallback, because the road has no name)

### DIFF
--- a/voice/de/de_tts.js
+++ b/voice/de/de_tts.js
@@ -409,12 +409,25 @@ function turn_street(streetName) {
 }
 
 function assemble_street_name(streetName) {
+	var toRef = streetName["toRef"]
+	/* Major roads highway=(motorway|primary|secondary|tertiary) in Germany 
+	 * have ref tags containing a type letter (A|B|L|K) and number, 
+	 * e.g. "ref=K 673" for a tertiary road. But only refs for motorways and highway=primary are
+	 * publicly visible on signposts, so reading the ref value of secondary or tertiary roads out loud
+	 * does not benefit the driver, as the information can not be cross-referenced with the
+	 * visible signs.
+	 * So only reference the roads by their ref attribute, if no actual street name is available,
+	 * which mainly happens in rural regions outside of towns.
+	 */
+	if ((toRef.startsWith("L") || toRef.startsWith("K")) && streetName["toStreetName"] != "") {
+		toRef = ""
+	}
 	if (streetName["toDest"] === "") {
-		return streetName["toRef"] + " " + streetName["toStreetName"];
-	} else if (streetName["toRef"] === "") {
+		return toRef + " " + streetName["toStreetName"];
+	} else if (toRef === "") {
 		return streetName["toStreetName"] + " " + dictionary["toward"] + " " + streetName["toDest"];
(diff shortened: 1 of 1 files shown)


vshcherb:
I would ask @sonora to review it. Though it's only german language (not generic) but probably it's worth it and enough

luziferius:
I probably should have opened it as a draft PR.

In the current form, it has some issues:

- Non-German locales still get the unnecessary ref read out
- It may suppress telling the ref outside of Germany for German users in unexpected spots. For example, if e.g. India were to use refs starting with K/L for motorways, a German visiting India would miss them.

I created this by hacking the json program and testing it on my phone as a proof of concept. I only used object fields known by looking at the usage in other methods. I haven’t looked how and where that `streetName` object is defined and which other fields it has.

Does that `streetName` object contain the country the street is in? Then it can be filtered for German roads and applied to all languages, which I’m willing to do.

sonora:
From what I see, this is not language but region-specific. And hence, if we wanted to implement this, it should not be done in the tts command files (language dimension) but in OsmAnd/VoiceRouter code, either via a manual or region-sensitive switch.

aceman444:
@sonora true, the problem is limited to region of Germany, the improvement should not be used outside Germany for all users which happen to select German voice. Maybe there is no better voice for them, but they will be using it outside Germany.

sonora:
From what I see, it is likely not a general improvement for quite a few use cases,a s per the comments above.

jfburke3:
Just commenting on this closed issue that in the U.S., almost every state highway is signed with the ref number, whether it has a name or not, so they should be read out.

sonora:
Very true, thanks for the comment. So basically, as I noted elsewhere, whether to read out these numbers would rather have to be made dependent on the country you are in, not the language you selected for your voice prompts.

PR #769, opened 2022-01-17
enable noexit=yes at lower level (14)

cmoffroad:
Identifying dead-end at a lower zoom level (14 instead of 16) would be very useful for ground survey planning.

### DIFF
--- a/obf_creation/rendering_types.xml
+++ b/obf_creation/rendering_types.xml
@@ -1741,7 +1741,7 @@
 		<entity_convert pattern="tag_transform" from_tag="parking_space" from_value="disabled" if_not_tag1="amenity" if_not_value1="parking_space" to_tag1="disabled" to_value1="yes" to_tag2="amenity" to_value2="parking_space"/>
 		<entity_convert pattern="tag_transform" from_tag="wheelchair" from_value="yes" if_tag1="amenity" if_value1="parking_space" to_tag1="disabled" to_value1="yes"/>
 
-		<type tag="noexit" value="yes" minzoom="16"/>
+		<type tag="noexit" value="yes" minzoom="14"/>
 		<type tag="highway" value="motorway_junction" minzoom="13" />
 		<type tag="junction" value="yes" minzoom="13"/>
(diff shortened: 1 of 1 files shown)


maraf24:
This may generate a lot of visual noise.
I often see noexit=yes placed on a road which is dead end only for cars, not for pedestrians or cyclists.
This is augmented by QA tools, which complain about two roads too close to each other - so mappers silence the warning with noexit=yes for both roads.

cmoffroad:
> This may generate a lot of visual noise.

It won't because the few built-in OsmAnd rendering styles that actually support noexit=yes are configured to render them only from zoom 16 or 18. So enabling noexit=yes starting at zoom=14 in the OBF creation process will have currently no effect for standard users.

For the ground survey plugin I built, It would however be very valuable for users to identify faster dead-end roads/paths that do not require further attention.  Other useful information like gates (barrier=\*) or fixme=\* on nodes are already available at zoom 14, so why not for noexit=yes?

> I often see noexit=yes placed on a road which is dead end only for cars, not for pedestrians or cyclists.
This is augmented by QA tools, which complain about two roads too close to each other - so mappers silence the warning with noexit=yes for both roads.

how does hiding quality issues makes maps better?  this would be one more reason why I would like users to see these issues so they can be fixed.

vshcherb:
This is quite limit

cmoffroad:
> This is quite limit

@vshcherb what is your exact argument for closing this request? 
It cannot be related to visuals as explained above, so is your concern about the increase in OBF size or something else I am not aware of ?

vshcherb:
Sorry... my message was in complete.

I wanted to say, that would overload the map too much and I think it should be implemented as a custom map style. 

If you want to just add this data to OBF, that could be done we just need to check how much impact it will have on average map size.

cmoffroad:
@vshcherb Ok. Yes, I would only need the data added to the OBF (as per the pull request).

I already maintain my own custom map style via a public plugin:
https://github.com/cmoffroad/osmand-offroad-survey-plugin

vshcherb:
@cmoffroad Please send then PR and if you can say impact in % or as KB on Germany or any other big country that would be great.

PR #770, opened 2022-01-22
OBF only: enable noexit=yes at lower level (14) 

cmoffroad:
**Important:** 

- This pull request only affects the OBF generation process by including `noexit=yes` starting at zoom level 14 instead of 16.
- It will not affect any built-in map styles, because those concerned only render `noexit=yes` at zoom level 16 and 18.

**Goal:** 

Identifying dead-end at a lower zoom level in this custom ground survey plugin:
https://github.com/cmoffroad/osmand-offroad-survey-plugin

**Estimated Size increase: + 0.9%**

Based on local tests done with Germany Bayern and Berlin states.


vshcherb:
hm.. 
```
Estimated Size increase: + 0.9%
```
it seems to be much bigger than I expected. Can you please post binary inspector outputs, it's not clear which 0.9% is increased (a map layer or a whole map)

PR #778, opened 2022-02-09
add tests

alisa911:


test-resources/test_turn_lanes.json
```diff
@@ -640,7 +640,7 @@
       "longitude": 35.43970754742625
     },
     "expectedResults": {
-      "32486314": "C|C|+TSLR"
```
vshcherb, line 643: "32486314:0": "C|C|+TSLR" - should stay? it's independent test case

PR #781, opened 2022-02-16
Fix avoid footways with test

Dima-1:


routing/routing.xml
```diff
@@ -1445,6 +1439,12 @@
 				<select value="0.2" t="highway" v="path"/>
 			</if>
 
+			<if param="avoid_footways">
```
vshcherb, line 1442: Same problem again... it will occur on highway=footway / surface=asphalt.
Change not structural.

PR #782, opened 2022-02-17
Observe maxwidth:physical limit on roads

aceman444:
Make the car profile observe maxwidth:physical tag set on roads and do not allow too narrow roads. See https://wiki.openstreetmap.org/wiki/Key:maxwidth:physical .

Will this work with the variable as "$maxwidth:physical", with the ":" character, which may have special effect in the declaration?


vshcherb:
Probably it's easier to test somehow.

aceman444:
There is a sample barrier at https://www.openstreetmap.org/node/9508004170 .

PR #783, opened 2022-02-22
Long road shields and park rendering fix

32Vache:
Added length 7, 8 and 9 for yellow square shields: In France, some refs of D roads (yellow shield) can be long. Fixed it so that a shield shows correctly on those:
 

Also reordered parks to show natural polygons on top for cases where parks are mapped in detail:


xmd5a2:
Road shields section merged manually. For reordering please create separate PR, but it needs testing in various places because changing order of `natural=water` and other often used tags can lead to unexpected issues.

PR #826, opened 2022-05-25
Equal treatment for all non-active railways

Famlam:
To my surprise, railways that are almost completely gone (abandoned=no more rails) or not in use (disused) are rendered at zoom levels up to 14, while a railway that is being constructed (i.e. at least as visible as abandoned, or more, in real life) or preserved (i.e. similar or better than "disused" railways) are hidden until zoom level 19.
This makes the treatment of all non-active railways equal

### DIFF
--- a/rendering_styles/default.render.xml
+++ b/rendering_styles/default.render.xml
@@ -3838,9 +3838,11 @@
 							<apply_if additional="layer=3" order="167"/>
 						</switch>
 						<switch>
+							<case tag="railway" value="abandoned"/>
 							<case tag="railway" value="construction"/>
+							<case tag="railway" value="disused"/>
 							<case tag="railway" value="preserved"/>
-							<apply_if moreDetailed="false" maxzoom="18" order="-1"/>
+                            <apply_if moreDetailed="false" maxzoom="18" order="-1"/>
 						</switch>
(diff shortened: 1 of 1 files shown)


maraf24:
Now you are treating disused rails the same as abandoned.
Disused rails are often ready to be used, just no trains are currently using them. This may change in one day.

Famlam:
True, but that is already the case anyway, see i.e.

https://github.com/osmandapp/OsmAnd-resources/blob/21ae640d5a672f4fca809aa545281f189edae616/rendering_styles/default.render.xml#L5098-L5099

https://github.com/osmandapp/OsmAnd-resources/blob/21ae640d5a672f4fca809aa545281f189edae616/rendering_styles/default.render.xml#L12053-L12054

Changing that would be a separate issue. (Although I would argue that both disused and abandoned are not strictly needed to be shown at low zoom levels, as in practice you'll not use them for navigation)

Famlam:
I hope the new patch applied resolves your objections by making existing infrastructure more prominent than nearly-gone infrastructure?

maraf24:
I reiterate in different words:
railway=disused are almost as important as railway=rail and should be rendered at the same zoom level (or almost the same).

PR #830, opened 2022-06-06
add destination_priority attr

alisa911:


routing/routing.xml
```diff
@@ -2436,6 +2452,10 @@
 			<select value="0.7"/>
 		</way>
 
+		<way attribute="destination_priority">
```
vshcherb, line 2455: needs to work without section somehow

PR #832, opened 2022-06-08
Consider cyclist and pedestrian hazards in routing

jeysal:
I recently learnt about the `hazard` tag. This adds some penalty for routing via a few types of hazardous places to bicycle and pedestrian routing. Cyclists and pedestrians are usually the parties negatively affected by these types of hazards. The penalty may in some cases result in hazards being avoided if an alternative route of otherwise similar quality is available.
It can be debated whether these hazards constitute "obstacle"s in the routing sense. But often for example a hazard=cyclists will mean merging onto a high-traffic road from a shoulder or kerb that may or may not be otherwise mapped and captured by routing, or hazard=dangerous_junction often will mean crossing a high-traffic road, both of which may require time to wait. Also, I only touched `obstacle` and left `obstacle_time` untouched.
Expansion of the routing file to include other hazards is of course possible, but this could be a start.

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -1665,6 +1665,9 @@
 			Any combination of values that we put here (into the routing.xml)
 			leads to routing errors. The above is only a compromise to minimize
 			those errors.-->
+            
+            <select value="5" t="hazard" v="cyclists"/>
+            <select value="10" t="hazard" v="dangerous_junction"/>
 			
 			<select value="9" t="crossing" v="traffic_signals"/>
 			<select value="1" t="crossing" v="uncontrolled"/>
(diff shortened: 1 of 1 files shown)


vshcherb:
Penalty is rather small I doubt a lot it could affect the routing, it could affect routing time but still not much. Please think of routing that it should impact any route if it doesn't impact it doesn't worth to make calculation more complicated cause it all takes time.

jeysal:
If the lookup for the select value is not constant time, i.e. there is a significant impact on routing time for each select tag (for example, the `<select>` tags are being looped over) as you imply then I would agree it is not worth it since the tag is too rare. You'll know this better than me. In that case not being able to use many `select`s in `routing.xml` is a pretty limiting factor on routing quality, but that's something to be solved in the routing engine, which is being re-written anyway I guess. Closing

PR #838, opened 2022-07-20
Edited Turkish TTS strings

MetehanOzyurek:
Signed-off-by: Metehan Özyürek <metehanc8s9@hotmail.com>

Edited TTS strings translations. Now translations are better;)


MetehanOzyurek:
@vshcherb will you upload new sound files according to new strings?

vshcherb:
Nope, TTS is automatically enabled. Recorded voice is not controlled by TTS, it needs to be recorded again

PR #850, opened 2022-09-26
Respect use_sidepath for bicycle, foot and horse

JeroenVanOort:
When `use_sidepath` is in use, routing should not be an option for the specific traffic. I don't know if this breaks any tests or if tests should be added to assert this behaviour.


vshcherb:
Please check latest night build in the areas you are familiar with this tag (you can find it with https://overpass-turbo.eu/)

PR #901, opened 2022-11-11
Add zh pack_rec_voice_std.sh

sonora:


vshcherb:
Added right so you can commit directly to master

sonora:
Great, thanks! I have tested some the last few hours, and think the new language packs I had uploaded are a clear improvement for all languages.

Beyond this, I guess we should no exaggerate our support for recorded voices, because the return of invest vs. functionality is not really high ... ;)

PR #908, opened 2022-11-26
TTS-PL fixes street/exit names in 3.11 3.12 4.3 5.2 5.3

akotynski:
fixes: https://github.com/osmandapp/OsmAnd/issues/7570#issuecomment-907590326
tested on emulator using Google TTS

3.4
There is:
Za około **trzech** **kilometrów** skręć w prawo w SR **osiemdziesięciu** w kierunku Rome.
Expected
Za około **trzy** **kilometry** skręć w prawo w SR **osiemdziesiąt** w kierunku Rome.

3.5 (Route 23)
There is:
Za trzysta pięćdziesiąt metrów skręć w lekko w prawo w Route **dwudziestu** trzech Main Street, następnie trzymaj się lewej strony.
Expected:
Za trzysta pięćdziesiąt metrów skręć w lekko w prawo w Route **dwadzieścia** trzy Main Street, następnie trzymaj się lewej strony.

3.11 (to exit 260B)
There is:
... w zjazd **dwustu sześćdziesięciu** B
Expected:
... w zjazd **dwieście sześćdziesiąt** B

3.12 (to exit 15B)
There is:
... w zjazd **piętnastu** B
Expected:
... w zjazd **piętnaście** B

5.2 (I 15)
There is:
Za czterysta pięćdziesiąt metrów wjedź na rondo i wybierz pierwszy zjazd w **"I piętnastu"** w kierunku Los Angeles
Expected:
Za czterysta pięćdziesiąt metrów wjedź na rondo i wybierz pierwszy zjazd w **"I piętnaście"** w kierunku Los Angeles

5.3 (Highway 60)
There is:
Wybierz drugi zjazd w Highway **sześćdziesięciu**.
Expected:
Wybierz drugi zjazd w Highway **sześćdziesiąt**.

### DIFF
--- a/voice/pl/pl_tts.js
+++ b/voice/pl/pl_tts.js
@@ -116,9 +116,9 @@ function populateDictionary(tts) {
 	dictionary["meters"] = tts ? "m" : "meters.ogg";
 	dictionary["around_1_kilometer"] = tts ? "około jeden kilometr" : "around_1_kilometer.ogg";
 	dictionary["around"] = tts ? "około" : "around.ogg";
-	//dictionary["kilometers"] = tts ? "kilometrów" : "kilometers.ogg";
-	dictionary["kilometers"] = tts ? "km" : "kilometers.ogg";
-	
+	dictionary["kilometers"] = tts ? "kilometrów" : "kilometers.ogg";
+	dictionary["kilometers_2-4"] = tts ? "kilometry" : "kilometers.ogg";
+
 	dictionary["feet"] = tts ? "stóp" : "feet.ogg";
 	dictionary["1_tenth_of_a_mile"] = tts ? "jedną dziesiątą mili" : "1_tenth_of_a_mile.ogg";
 	dictionary["tenths_of_a_mile"] = tts ? "dziesiątych mili" : "tenths_of_a_mile.ogg";
@@ -162,9 +162,9 @@ function num_fe_str(number) {
 	return (numberWithoutOnes ? numberWithoutOnes.toString() + " " : "") + onesPhrase;
 }
 
-function time_str(time, pluralTimeForm, pluralTimeFormBetweenTwoAndFour) {
-	var ones = time % 10;
-	var tens = time % 100 - ones;
+function plural_str_form(value, pluralTimeForm, pluralTimeFormBetweenTwoAndFour) {
+	var ones = value % 10;
+	var tens = value % 100 - ones;
 	if (tens === 10) {
 		return pluralTimeForm;
 	}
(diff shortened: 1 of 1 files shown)


FreddieChopin:
If I read that correctly, the problem with wrong numerals should be fixed, while it is definitely not (I even reported that some years ago in osmandapp/OsmAnd#6088).

When the navigation wants me to turn to highway "1, 91" (highway numbers, Polish "droga krajowa" etc.) it is pronounced as "first, ninetyone" ("skręć w pierwszą, dziewięćdziesiąt jeden", while it should be "one, ninetyone" ("skręć w jeden, dziewięćdziesiąt jeden").

It is obviously a problem with TTS which just gets fed the numbers, as it generally affects only the first number, not the later ones (where a road has multiple), so I was thinking that maybe it could be easily fixed by adding explicit "drogę numer ..." ("route number ...") before the number(s) could improve that?

It is super annoying...

FreddieChopin:
OK, I see what the "problem" now. The example prompts available in the development plugin always has some other word before the number, like "highway 60", "I 15", "Route 23" or "SR 80"). In Polish TTS there is no problem for "numbers" which have any prefix, like highway "A1" or express roads "S1". The problem is only for lower categories, like national roads, which are just numbers. So this PR fixes the cases in the devel plugin, but the "real-world" issues in Poland remain (;

@akotynski you seem like a person who understands the internals of OsmAnd's TTS, so maybe you could help with that issue?

akotynski:
Hey, I could help you with that. However, I suggest you create a new ticket instead of opening a new discussion in the existing PR. The old PR isn’t the appropriate place for such a discussion.

I’m not entirely clear on your case. Is this only an issue with the DK1 route? Could you please provide some routes in GPX format so that I can undestand the exact case?

FreddieChopin:
The best place for discussion will be https://github.com/osmandapp/OsmAnd/issues/6088 . I'll try to come up with some examples and post them there.

sonora:
Please also try different TTS engines and in their latest versions. This smells a bit of a bug in a certain engine which, if true, should be fixed in the engine, not in OsmAnd.

FreddieChopin:
@sonora This is with default Google's TTS engine, so I think something 99% of Osmand users in PL will use.

sonora:
1. Please open a separate issue to dsicuss this.
2. Perhaps you want to experiment with an app like "TTS Test" to find out under which circumstances exactly the issue occurs: Make sure your device settings are set to Polish. Then experiment with the comma and the blank, it seems that google's TTS engine may for some reason misinterpret the punctuation here for some reason...

PR #916, opened 2022-12-28
Removed an useless `;admin_level=6`

ghost:
Removed `;admin_level=6` that seems useless.


xmd5a2:
Why did you decided that admin_level=6 here is useless? It is used to search corresponding translation in downloaded OSM data. There can be more than 1 translations.

PR #928, opened 2023-01-15
Treat dog_toilet like dog_park

Famlam:
Currently dog_toilet is not rendered. Since it's a relatively rare key, I guess it makes sense to treat it as dog_park, even though there are actually subtle differences

https://wiki.openstreetmap.org/wiki/Tag:amenity%3Ddog_toilet


xmd5a2:
I think we need to add separate POI

Famlam:
Ok, then I'll close this PR, don't know how to do that

PR #932, opened 2023-01-20
Unify use of JS "equals"

sonora:
@vshcherb We use an arbitrary mix of "===" and "==" in our TTS .js config files, it seems that "==" suffices?


vshcherb:
"5" == 5 - true
"5" === 5 - is not true

So it's not clear what do we need probably ===

sonora:
Exactly my thinking. So we should merge it, right?

PR #934, opened 2023-01-29
Reword `hu` and `hu-formal` exit and arrival voices

MmAaXx500:
Current highway and roundabout voices are in a bad shape because words are not in the right order or makes no sense. To make them better, I have reworded highway and roundabout exits and added a colon after arrival voices.

I have tested the changes with the developer plugin and on some routes.

Ogg files not updated because I see no way to do that currently. This means that the prerecorded voices are untested and broken. (`exit2.ogg` is a duplicate of `exit1.ogg`)


MmAaXx500:
@vshcherb Should I make a pr for 4.3 or this only going to be released in 4.4?

vshcherb:
4.4 should start rolling out in 2 weeks so not that critcial

PR #935, opened 2023-02-06
A basic profile for mopeds

JeroenVanOort:
I've been working on a custom profile for mopeds for some time, but since some extra data is needed in the obf files (inferring speed from [`zone:traffic`](https://wiki.openstreetmap.org/wiki/Key:zone:traffic)) and I was making a PR for that, I figured I could just as well share the rest of my efforts too.

Some questions:

1. I see that `bicycle=use_sidepath` is [transformed](https://github.com/osmandapp/OsmAnd-resources/blob/2915920078e17fdd215557ffdfdde5bcd3f273fa/obf_creation/rendering_types.xml#L205) to `bicycle:no` (although that would make [this](https://github.com/osmandapp/OsmAnd-resources/blob/2915920078e17fdd215557ffdfdde5bcd3f273fa/routing/routing.xml#L1231) seem superfluous). Would this be a good thing to do for `moped=use_sidepath` too? It might help keep the data size down, a bit.
2. Is there any documentation available on `minSpeed`, `sharpTurn` and `slightTurn`? I have guessed the values, but I don't really understand what they do.
3. Would the new profile need a logo or translations? And do you have any guidance on how to add those?

I plan to add obstacles later. I wanted to get the basics working first.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -17,3 +17,6 @@ continuation_indent_size = 4
 [*.cmake]
 indent_style = tab
 continuation_indent_size = 4
+
+[*.xml]
+trim_trailing_whitespace = false
--- a/obf_creation/rendering_types.xml
+++ b/obf_creation/rendering_types.xml
@@ -578,6 +578,11 @@
 	<entity_convert pattern="tag_transform" from_tag="maxspeed:type" from_value="UZ:rural" to_tag1="maxspeed" to_value1="100" map="no" poi="no"/>
 	<entity_convert pattern="tag_transform" from_tag="maxspeed:type" from_value="UZ:urban" to_tag1="maxspeed" to_value1="70" map="no" poi="no"/>
 
+	<entity_convert pattern="tag_transform" from_tag="zone:traffic" from_value="NL:rural" to_tag1="maxspeed" to_value1="80" if_not_tag1="highway" if_not_value1="cycleway" map="no" poi="no"/>
+	<entity_convert pattern="tag_transform" from_tag="zone:traffic" from_value="NL:urban" to_tag1="maxspeed" to_value1="50" if_not_tag1="highway" if_not_value1="cycleway" map="no" poi="no"/>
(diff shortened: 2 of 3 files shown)


vshcherb:
Looks interesting... Though there is possibly no need to duplicate profile and moped could be done as variation of bicycle similar how 'truck' is part of 'driving'.

I think we need to create an issue  in OsmAnd keep track of that profile and other work to be done i.e. icon / profile name.

vshcherb:
1. probably yes
2. sharpTurn penalties for sharp / slight turns (angle is changing but possible to find in the code)

JeroenVanOort:
Thanks for your feedback!

1. Regarding making this a derived profile: that would be nice, but how would I set the `routingProfile` parameters like `maxSpeed` and `defaultSpeed`? That I can't find.
2. I've created an issue in the OsmAnd repository for adding the icon(s) and translation(s): https://github.com/osmandapp/OsmAnd/issues/16446
3. I've optimized the conversion of `use_sidepath`.
4. I'll leave `sharpTurn` and `slightTurn` for what they are right now. Can always tweak later.

vshcherb:
1. Good point. maxspeed  is a routing attribute so probably we can add support syntax similar like this
```<attribute name="minPointApproximation" value = "5" profiles="moped"/>```

However the question is always how much copy/paste - difference between bicycle and moped profile, if it shares < 50% code than it's a good to have separate, if > 90% better to build a shared support for it.

JeroenVanOort:
Adding support for such syntax seems too complex for me personally. May we can keep it simple for now and just have the separate profiles. If the need to have a derived profile emerges/remains, that work can be done later.

JeroenVanOort:
Can we get this merged? Would be great to have starting point!

vshcherb:
I don't think we're ready to  merge before we take full task with creating profile, so for now please leave it open. 

However let me check please.

PR #937, opened 2023-02-21
Add several recycling waste types

eunikolsky:
This adds four new types of recycling waste that I see in OSM. They are all present in the table for [`amenity=recycling`][table]. This also adds several translations for the names.

I've verified the changes locally after creating a piece of the map with these recycling POIs.

See also: https://github.com/osmandapp/OsmAnd-iOS/pull/2593.

[table]: https://wiki.openstreetmap.org/wiki/Tag:amenity=recycling#Accepts


vshcherb:
Yes it needs to be merged to main source android. https://github.com/osmandapp/OsmAnd-resources/blob/master/poi/phrases/README.txt

PR #962, opened 2023-04-21
Add OsmAndOwner to obf

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd-Issues/issues/1885

### DIFF
--- a/protos/OBF.proto
+++ b/protos/OBF.proto
@@ -707,7 +707,8 @@ message OsmAndRoutingIndex {
 }
 
 message OsmAndOwner {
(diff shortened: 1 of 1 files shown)


protos/OBF.proto
```diff
@@ -704,3 +706,8 @@ message OsmAndRoutingIndex {
 
 }
 
+message OsmAndOwner {
+	required string owner = 1;
```
vshcherb, line 710: Suggested change:
- required string name;
- optional string resource;
- optional string description;
- optional string pluginid;

PR #967, opened 2023-05-07
Esperanto TTS package

slashdevslashurandom:
Here's my attempt at an Esperanto TTS voice for OsmAnd. I tried to make the grammar work properly, but as I am a beginner, I may have made some mistakes. It should be usable with the [RHVoice](https://rhvoice.org/) TTS package, as it has an Esperanto voice.

### DIFF
--- a/voice/eo/eo_tts.js
+++ b/voice/eo/eo_tts.js
@@ -1,0 +1,600 @@
+// IMPLEMENTED (X) or MISSING ( ) FEATURES, (N/A) if not needed in this language:
+//
+// (X) Basic navigation prompts: route (re)calculated (with distance and time support), turns, roundabouts, u-turns, straight/follow, arrival
+// (X) Announce nearby point names (destination / intermediate / GPX waypoint / favorites / POI)
+// (X) Attention prompts: SPEED_CAMERA; SPEED_LIMIT; BORDER_CONTROL; RAILWAY; TRAFFIC_CALMING; TOLL_BOOTH; STOP; PEDESTRIAN; MAXIMUM; TUNNEL
+// (X) Other prompts: gps lost, off route, back to route
+// (X) Street name and prepositions (onto / on / to) and street destination (toward) support
+// (X) Distance unit support (meters / feet / yard)
+// (N/A) Special grammar: (please specify which)
+// (X) Support announcing highway exits
+
+var metricConst;
+var dictionary = {};
+var tts;
+
+//// STRINGS
+////////////////////////////////////////////////////////////////
+function populateDictionary(tts) {
+	// ROUTE CALCULATED
+	dictionary["route_is"] = tts ? "La kurso estas" : "route_is.ogg";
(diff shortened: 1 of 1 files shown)


vshcherb:
If @sonora could review it would be great

voice/eo/eo_tts.js
```diff
@@ -0,0 +1,660 @@
...
+			return dictionary["left_keep"];
+			break;
+		case "right_keep":
+			return dictionary["right_keep"];
+			break;
+	}
+}
+
+function then() {
+	return dictionary["then"] + " ";
```
sonora, line 365: Here you may want to test if better intonation requires a preceding comma, i.e. `return (tts ? ", " : " ") + dictionary["then"] + " ";` Most languages have this to better pause these concetated prompts.

sonora:
@slashdevslashurandom I think you did a good job! I see no flaws, and it should work ok. Thank you!

- I have two comments for you to check, see above.

@vshcherb Good to merge, I would say. Could wait for feedback on the two items I noted above.
Need to of course create the string and make the modification to include as resource in the apk, please let me know if you want me to take that on.

voice/eo/eo_tts.js
slashdevslashurandom: Yeah, that makes sense.

voice/eo/eo_tts.js
```diff
@@ -0,0 +1,660 @@
...
+}
+
+function turn_street(streetName) {
+	if ((streetName["toDest"] == "" && streetName["toStreetName"] == "" && streetName["toRef"] == "") || Object.keys(streetName).length == 0 || !tts) {
+		return "";
+	} else if (streetName["toStreetName"] == "" && streetName["toRef"] == "") {
+		return dictionary["toward"] + " " + streetName["toDest"];
+	} else if (streetName["toRef"] == streetName["fromRef"] && streetName["toStreetName"] == streetName["fromStreetName"]) {
+		return dictionary["on"] + " " + assemble_street_name(streetName);
+	} else if (streetName["toStreetName"] == "" && streetName["toRef"] == streetName["fromRef"]) {
```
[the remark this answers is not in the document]
slashdevslashurandom: Oh, that's because I actually started off by copying the Russian voice script at first, instead of the English one. Then I realized using the English one would be easier (and because the Russian script didn't actually have a solution for the problems I was anticipating)... I'll copy the English code, thanks.

slashdevslashurandom:
I made the suggested changes. I don't know if there are other files here that also need to be edited to make the package work, and ideally it would be nice if someone could record or generate audio files for a prerecorded version as well...

vshcherb:
should be available in night build for android

vshcherb:
Probably you are right it should be there to work however  it could be only for updates not for downloads

sonora:
Yes, you're right, it is for updates, I think. I have added it: https://github.com/osmandapp/OsmAnd/commit/c68680ed995b49180c083fb7af8a87b0a0291684

There was already a `lang_eo` string for the UI, so the menu voice selection menus works already.

PR #993, opened 2023-07-17
Stadium and playground color changed

Velomeloman:
@sonora, please look at this color changes. Stadium had same color as building. Playground color was too saturated.

### DIFF
--- a/rendering_styles/Touring-view_(more-contrast-and-details).render.xml
+++ b/rendering_styles/Touring-view_(more-contrast-and-details).render.xml
@@ -542,7 +542,7 @@
 	</renderingAttribute>
 	<renderingAttribute name="stadiumColor">
 		<case nightMode="true" attrColorValue="$buildingColorNight"/>
-		<case attrColorValue="#bca9a9"/>
+		<case attrColorValue="#bef7cc"/>
 	</renderingAttribute>
 	<renderingAttribute name="stadiumStrokeColor">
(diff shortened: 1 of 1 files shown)


sonora:
Thanks for the suggestion! I have the following thoughts:

Touring view is a rather 'busy' map style, and inflating the number of different colors used for areas is likely detrimental...

That's why for **stadium**, characteristically being a building (tiered seating at the very minimum) and a pitch (could even be covered by a roof) I use the regular building color. Same as for churches, too, no further diversification in the color dimension. Often, stadiums are recognizable by their unique shape, anyway.

In principle the same for **playground**: I am not against making this less intense, but playgrounds are often of other green backgrounds, so there needs to be some contrast still. And: The same color #33cc99 is used for for **sportsCentre**, so we would have to change in sync, to not create more colors. Need some time to investigate.

Velomeloman:
The leisure = stadium tag applicable for whole territory of the stadium. The building = stadium for building area. I want to change only whole area color similary to default osm render.
 The territory of the stadium has same color as all the buildings inside:
https://www.openstreetmap.org/relation/4229877/





Building = stadium has no regular building color: https://www.openstreetmap.org/relation/11033475/




Example with new colors of stadium and playground

sonora:
You are right, I had overlooked that - thank you.

I welcome the proposal and have merged it here: https://github.com/osmandapp/OsmAnd-resources/pull/995

PR #1001, opened 2023-08-06
Fix typo in beep voice packs

mikehgentry:
This is something of a test PR, so please let me know if I've done anything wrong.

It corrects the comparison operators later in the beep-complex scripts from < to <=, and should have no practical effect.

I note that it's also stripped a final line of whitespace from each file (perhaps because I've done it through github's web interface?)


sonora:
@mikehgentry Yes, final line seems a web interface artifact - no prob from what I see.

Regarding `<=`, I've checked the other ~50 _tts.js_ files, and all use `<` in that code block, so let's make it consistent and change the PR to rather replace `<=` by `<`?

PR #1003, opened 2023-08-07
wetland z-order=4

quantenschaum:
This would fix https://github.com/osmandapp/OsmAnd/issues/17835

Does it have negative side effects?

### DIFF
--- a/rendering_styles/nautical.render.xml
+++ b/rendering_styles/nautical.render.xml
@@ -557,7 +557,7 @@
 			<case tag="natural" value="crater"/>
 			<case tag="natural" value="hot_spring"/>
 			<case tag="natural" value="geyser"/>
-			<case tag="natural" value="wetland" order="6"/>
+			<case tag="natural" value="wetland" order="4"/>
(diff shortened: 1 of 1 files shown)


rendering_styles/nautical.render.xml
```diff
@@ -557,7 +557,7 @@
 			<case tag="natural" value="crater"/>
 			<case tag="natural" value="hot_spring"/>
 			<case tag="natural" value="geyser"/>
-			<case tag="natural" value="wetland" order="6"/>
```
vshcherb, line 560: orders should be avoid using at all cause it's not consistent with default osm styles, but here there was exception unclear why

quantenschaum: But with no order attribute set wetland polygons get (partially) displayed on top of cable areas for example, this does not look good.

vshcherb: Z-fighting... on sea there are different rules than on the land, so I'm not sure @xmd5a2

PR #1004, opened 2023-08-07
Change comparison operators in beep voice packs for consistency 

mikehgentry:
@sonora 

Sorry, not sure that was the right way to go about switching it round to the other way, but this seems to have worked :)

But yeah, it makes sense to have it consistent with all the other checks.

Thanks for your assistance.

ref #1001


sonora:
Cool  thanks! It wil update in the downloads as available from within the app momentarily!

PR #1005, opened 2023-08-17
Fix bicycle routing over private access obstacles

Velomeloman:


Velomeloman:
Sorry for the extra commits in the pull request. I am new to github. In fact, there is one added line that fixes the routing error.

PR #1020, opened 2023-11-05
Soften height_obstacles to fix issue and speed up routing (Issue #18290)

RZR-UA:
Formula proposed:

**percent (incline factor) * difficulty (factor) = xml_penalty_value**

1.03 * **1.5** = 1.5
1.09 * **2.5** = 2.5
1.17 * **5.5** = 6.4
1.25 * **20** = 25.0

highway=**path** gains penalty by 1.2 (**+20%**)

Actual final penalty formula is: diff (segment difference in meters) * xml_penalty_value (_formula kept as is_)

Tested (and fixed) on issue #18290

Tested on familiar terrain of Carpatian mountains (bicycle).

The issue: https://github.com/osmandapp/OsmAnd/issues/18290

Results:

base routing_cost=6800 (w/o height_obstacles)

Before proposed formula: routing_cost=115000 (16x), native lib hangs with timeout

**After proposed formula: routing_cost=22626 (3x), native lib seems OK**

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -1433,24 +1433,24 @@
 					<select value="-1" t="highway" v="track"/>
 				</gt>
 				<gt value1="-25" value2=":incline">
-					<select value="180" t="osmand_highway_integrity_brouting_low" v="yes"/>
-					<select value="180" t="highway" v="path"/>
-					<select value="170" t="highway" v="track"/>
-				</gt>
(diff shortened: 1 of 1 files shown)


xmd5a2:
This PR works as expected and significantly reduces routing time and visited segments in mountains. The differences are insignificant on less hilly relief.

PR #1022, opened 2023-11-14
[DELETE - refactor need] Add routing test 63 [ignored but HH-branch-ok] (Issue #18093)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/18093

PR #1027, opened 2023-12-10
consider good surface types as well

colorimeter:
Consider the road quality differences, even if they are good.

### DIFF
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -881,6 +881,10 @@
 				<select t="hgv" v="delivery" value="0.15"/>
 			</if>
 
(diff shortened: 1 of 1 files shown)


routing/routing.xml
```diff
@@ -881,6 +881,10 @@
 				<select t="hgv" v="delivery" value="0.15"/>
 			</if>
 
+            <select value="1.1" t="smoothness" v="excellent"/>
```
vshcherb, line 884: They will override all other settings, so here we mention on top only disqualifying settings like < 0.4. 

Please Keep in mind all the values are not multiplied but only the top one is selected

PR #1030, opened 2023-12-29
Prepare SVG icons for OpenGL osmandapp/OsmAnd#16920

cepprice:


goldfndr:
osmc_brown_diamond_left wasn't included and is no longer rendering on a map visualization (for guard rails) I have within <line> (excerpted):

<case tag="barrier" value="fence">
<apply_if minzoom="16" pathIcon="osmc_brown_diamond_left" pathIconStep="14"/>
<apply_if minzoom="16" nightMode="true" pathIcon="osmc_orange_diamond_left"/>
</case>

Please add it.

PR #1031, opened 2024-01-05
Properly render ways with piste:type=* containing ;-separated values

broadwaylamb:
For example, [this way](https://www.openstreetmap.org/way/158183374) mapped with the `piste:type=nordic;fatbike` tag was not displayed on the ski map, because the rendering config did not support this set of values. [OpenSnowMap](http://opensnowmap.org), however, displays it correctly, with both ski and fatbike icons.

The support is added here. Such ways are rendered as nordic ski tracks.

Also, if a piste is both `nordic` track and `hike`, we render caps on it like for `hike` pistes (see the screenshot).
Screenshot 2024-01-05 at 21 00 55

### DIFF
--- a/obf_creation/rendering_types.xml
+++ b/obf_creation/rendering_types.xml
@@ -2547,9 +2547,18 @@
 		<entity_convert pattern="tag_transform" from_tag="piste:type" from_value="downhill" if_not_tag1="piste:oneway" if_not_tag2="oneway" if_not_value2="no" if_not_tag3="aerialway" to_tag1="piste:type" to_value1="downhill" to_tag2="piste:oneway" to_value2="yes" apply_to="way"/>
 		<type tag="piste:type" value="nordic" minzoom="9" />
 		<type tag="piste:type" value="skitour" minzoom="9" relation="true"/>
+        <type tag="piste_hiking_allowed" additional="true" order="90" notosm="true"/>
 		<entity_convert pattern="tag_transform" from_tag="piste:type" from_value="skitour" if_not_tag1="piste:grooming" to_tag1="piste:type" to_value1="skitour" to_tag2="piste:grooming" to_value2="backcountry"/>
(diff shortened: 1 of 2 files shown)


obf_creation/rendering_types.xml
```diff
@@ -2547,9 +2547,18 @@
...
+        <entity_convert pattern="tag_transform" from_tag="piste:type" from_value="hike;fatbike;nordic" to_tag1="piste:type" to_value1="nordic" to_tag2="piste_hiking_allowed" to_value2="yes"/>
+        <entity_convert pattern="tag_transform" from_tag="piste:type" from_value="fatbike;nordic;hike" to_tag1="piste:type" to_value1="nordic" to_tag2="piste_hiking_allowed" to_value2="yes"/>
+        <entity_convert pattern="tag_transform" from_tag="piste:type" from_value="nordic;hike;fatbike" to_tag1="piste:type" to_value1="nordic" to_tag2="piste_hiking_allowed" to_value2="yes"/>
+        <entity_convert pattern="tag_transform" from_tag="piste:type" from_value="hike;nordic;fatbike" to_tag1="piste:type" to_value1="nordic" to_tag2="piste_hiking_allowed" to_value2="yes"/>
+        <entity_convert pattern="tag_transform" from_tag="piste:type" from_value="nordic;fatbike;hike" to_tag1="piste:type" to_value1="nordic" to_tag2="piste_hiking_allowed" to_value2="yes"/>
+        <entity_convert pattern="tag_transform" from_tag="piste:type" from_value="fatbike;hike;nordic" to_tag1="piste:type" to_value1="nordic" to_tag2="piste_hiking_allowed" to_value2="yes"/>
```
broadwaylamb, line 2561: Here I enumerate all the permutations of these three values. I wonder if there is a more robust way to do this.

xmd5a2:
Thanks!

PR #1037, opened 2024-01-15
Add test for "Incorrect turn lanes #18610"

Dima-1:


vshcherb:
Implemented slightly different approach with merged lanes

PR #1040, opened 2024-01-29
Increase walkRadius / walkChangeRadius for public_transport

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/18809

PR #1048, opened 2024-02-22
added music venue

FabiRich:
just wanted to add the amenity music_venue, because I could not find the place of it in OSMAnd. Here is the documentation https://wiki.openstreetmap.org/wiki/Tag:amenity=music_venue


dmpr0:
Added available in March maps

PR #1065, opened 2024-04-07
Add LT voice, generated using Google TTS with adjustments

embar-:


vshcherb:
For that request we need issue to be created cause it needs to be packaged and uploaded later to servers

embar-:
OK, issue created: https://github.com/osmandapp/OsmAnd/issues/19502

PR #1077, opened 2024-05-24
Import Nautical-based Marine style by Adam (quantenschaum), add light sectors to Nautical style

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/16894

PR #1090, opened 2024-07-12
Fix lt_tts.js by replacing "let" into "var"

embar-:
Otherwise Lithuanian TTS may not work in OsmAnd

### DIFF
--- a/voice/lt/lt_tts.js
+++ b/voice/lt/lt_tts.js
@@ -664,9 +664,9 @@ function modify_street_name(street_name, grm_case) {
 	while (street_name.length > 3 && /[A-Z]/.test(street_name.charAt(0)) && street_name.substring(1, 3) == ". " && street_name.endsWith(' g.'))
 		street_name = street_name.substring(3);
     // remove initials in middle of street name, e.g. remove middle "S" from "S. Dariaus ir S. Girėno g."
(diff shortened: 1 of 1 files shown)


embar-:
Thanks. Can you please rebuild Lithuanian TTS voice package and 
distribute it? The fixed version is still not visible via OsmAnd mobile 
application...

vshcherb:
Migrated to 4.8 branch, nightly 4.9 is available already

embar-:
I see now that language packages are different for OsmAnd versions...
Thanks for backporting to v4.8.
Can you please backport this fix to v4.7 also?

vshcherb:
Nope 4.7 is frozen for us and we're not going to produce any builds any more

embar-:
We need re-package LT voice package with "recorded OGG audio files", not only "single TTS" package.

vshcherb:
@sonora will later do it

sonora:
Done.

embar-:
2024-08-01 18:40, Hardy rašė:
> Done.


I just opened OsmAnd 4.8 and still don't see update for Lithuanian voice.

I also downloaded 
https://download.osmand.net/download?standard=yes&file=lt_0.voice.zip 
and it's date is 2024-05-08 - outdated.

PR #1092, opened 2024-07-15
Add turn-lanes-test 85 (Issue #20262)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/20262

PR #1093, opened 2024-07-15
enable additional="true" for ford=yes so that ford highways can be rendered differently

cmoffroad:
According to the OpenStreetMap wiki (https://wiki.openstreetmap.org/wiki/Key:ford#On_a_way), the tag `ford=yes` can be applied to highways that cross long riverbeds, which are commonly found worldwide.

At present, only nodes tagged with `ford=yes` are rendered with a specific icon. Applying `ford=yes` to a highway tag doesn't change its appearance.

To enhance rendering support for `highway=* + ford=yes`, I propose to enable `additional="true"` on the related rule.

### DIFF
--- a/obf_creation/rendering_types.xml
+++ b/obf_creation/rendering_types.xml
@@ -2078,7 +2078,7 @@
 		<type tag="area:highway" value="cycleway" minzoom="16" order="10"/>
 		<type tag="area:highway" value="track" minzoom="16" order="10"/>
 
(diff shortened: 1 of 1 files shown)


xmd5a2:
We already render *highway=* + *ford=yes* with an icon. Also semitransparent blue line is rendered but only on tracks and lower. I've [fixed this](https://github.com/osmandapp/OsmAnd-resources/commit/1e2afee805d98180b7118771fe10fc7a4954cb12).
изображение

PR #1094, opened 2024-07-16
Unify inner_download_prefix to name for "nordrhein-westfalen"

Henning-Lenz:
All other regions which have sub-regions keep their region name as `inner_download_prefix` if it contains a dash "-" (e.g. "canada_british-columbia", "france_auvergne-rhone-alpes" and "france_great-east"). Therefore it seems reasonable to also apply that to "germany_nordrhein-westfalen" instead of introducing an underscore (which interferes with splitting download names correctly).


xmd5a2:
Merged manually

PR #1097, opened 2024-07-29
Add Approximation tests round03, round04 (also enable round01, round02)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/20280

PR #1100, opened 2024-07-30
Add tests for Issue #19773 vehicle:conditional=no

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/19773

test-resources/test_routing.json
```diff
@@ -1229,5 +1229,49 @@
       "23487067": false,
       "23487607": true
     }
+  },
+  {
+    "testName": "68. vehicle:conditional=no https://github.com/osmandapp/OsmAnd/issues/19773 (1)",
+    "params": {
+      "vehicle": "car",
+      "minRouteLength": 140,
```
vshcherb, line 1237: no need to support weird minroutelength... 
We should proper support "expectedResults":  if needed points then points, so it's easier to understand what's the test case
```
"4692930:1": true,
"4692930:8": false
```

RZR-UA: improved (slightly other way)

vshcherb:
We need to store osm.gz as well, in a year osm will change and we won't be able to understand node details properly whether it was osm -> obf or obf -> osm glitch if we want to debug and open josm code. Now it doesn't matter as it possible to check OSM but later it will be difficult

RZR-UA:
Added

PR #1102, opened 2024-07-31
Show emergency access points starting from 17 zoom

Andrewschenko:


vshcherb:
We should keep 18th zoom, so the issue is somewhere else as on screenshots I see 15-16 zoom

PR #1110, opened 2024-08-19
Add current OSM source for Routing_test_68_69

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/19773#event-13922237935

PR #1111, opened 2024-08-19
Add vice-versa tests for vehicle:conditional=no (71, 72)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/19773

PR #1114, opened 2024-08-27
Test synthetic lanes for traffic rules compliance

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2639

PR #1127, opened 2024-09-21
Touring-view: fix display issue for bicycle=desginated in bike profile

sonora:


Velomeloman:
I suggest setting a value pathEffect_5 = "8_4" to unify the stroke length for cycleways in the line #1306.

sonora:
Well, the scheme has always been that 4_2 is the stroke pattern for dedicated footway/cycleway/bridleway, while 8_4 is the pattern for ways which have bicycle as a side/additional tagging (designation, lane, yes, etc.)...

Velomeloman:
Additional tags are well expressed by the strokeWidth. In my opinion, the same stroke length is well associated with the same purpose of the line. Short strokes are little bit confusing. In other styles, all cycleways have the same stroke length. It's also more visually pleasing.

sonora:
Agreed - let's try.

PR #1128, opened 2024-09-21
Touring view: small improvements

sonora:


Velomeloman:
Hi. I wrapped unifying of (foot-) path/footway color into Alpine Hiking = "true" condition. Because the differences between a path and a footway are important in "citizen" mode.




Line 235:
```
	<!-- Path color used is by profile. Unify (foot-)path/footway. Also avoids SAC color conflicts. -->
	<apply_if alpineHiking="true">
		<renderingAttribute name="footwayColor">
			<case attrColorValue="$pathColor">
				<apply_if nightMode="true" attrColorValue="$pathColorNight"/>
			</case>
		</renderingAttribute>
	</apply_if>
```

Velomeloman:
But in pedestrian profile, the footways are still black. I didn't understand how to fix it.

sonora:
Users frequently requested this simplification to reduce color clutter and focus on more significant distinctions, so I guess it is here to stay.

If you are trying to manually circumvent it, I guess you need to move the apply_if inside the rendering attribute, assign a separate color there per condition as needed.

PR #1137, opened 2024-09-24
Add examples

vshcherb:


rendering_styles/default.render.xml
```diff
@@ -65,8 +65,12 @@
 		ignorePolygonArea - process polygon even if it is filtered by area
 		ignorePolygonAsPointArea - process polygon as point (center) even if it is filtered by area
 	-->
+	<!-- rendering classes, allow programatically change appearance (association) -->
+	<renderingClass name="access" values="toll,private,destination,..." attributes="zoom">
```
vshcherb, line 69: renderingExample="highway=secondary,access=$value"

vshcherb: rClass="route.cycling.network.$network; route.cycling.type.$type"

PR #1148, opened 2024-10-25
Actualize OSM routes types, use activities icons, remove useless prefixes, add skates and fitness to other activities

RZR-UA:


RZR-UA:

xmd5a2 (approved):

PR #1149, opened 2024-11-07
add tags

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2708

goldfndr:
"velo" is in twice in activities.json.

goldfndr:
"velo" is in twice in activities.json.

PR #1154, opened 2024-11-14
Add osmand_poi_lat_lon

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/8159


vshcherb:
Only master needs to be for maps creation

PR #1159, opened 2024-11-24
#6088 force TTS to use correct PL number forms in street name

akotynski:
solves [https://github.com/osmandapp/OsmAnd/issues/6088](https://github.com/osmandapp/OsmAnd/issues/6088)


vshcherb:
Please test it

FreddieChopin:
Tested a downloaded version of the pl_tts file and it finally sounds correct (;

PR #1160, opened 2024-11-25
Removed roboto font files

Corwin-Kh:


Corwin-Kh:
No need to remove these resources at the moment. The problem was resolved by cleaning build source files

PR #1169, opened 2024-12-19
TravelGpx: add url_text as a child of url (support gpx <link><text>)

RZR-UA:


RZR-UA:

PR #1171, opened 2025-01-06
Add test 74 (HH-routing via roadblock)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/21574

PR #1172, opened 2025-01-13
[draft] Add activity type tags for ClickableWay feature; ski route fixes (later)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2846

PR #1187, opened 2025-02-27
Update route_type, route_bbox_radius, route_shortlink_tiles tags

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/16808

PR #1189, opened 2025-03-05
Fix OsmGpxWriteContext regression (avoid route_type for Nodes)

RZR-UA:


RZR-UA:

PR #1193, opened 2025-03-17
Add order="91" to all route_activity_type(s)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/16808

PR #1198, opened 2025-03-24
Fix duplicate poi types after request #1196

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/21972

PR #1202, opened 2025-03-28
Fix naming issue in regions.xml for Iran provinces 

mokazemi:
Name of two of Iran provinces should have been swapped.

"Kerman" and "Kermanshah" are two different provinces and shouldn't be mistaken with each other.


xmd5a2:
Already fixed

PR #1215, opened 2025-05-01
[master] Support oneway:motorcycle in motorcycle routing

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/21940

PR #1221, opened 2025-05-19
[5.0] Hide internal tag, fix useless multimenu for dotted "main entrance"

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/22579

PR #1223, opened 2025-05-20
[master] Hide Travel Article tags in Amenity Menu

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/21954

PR #1231, opened 2025-07-09
Workflow test dummy change

matthijskooijman:
This is a test PR that serves to showcase the github workflow I am working on. This PR makes a single change to the rendering style (making cycle route node text red instead of blue), which should produce a single OSF file containing just `routes.addon.render.xml` and an `items.json` containing it.


matthijskooijman:
W00ps. submitted this to the wrong repo, sorry for the noise...

PR #1236, opened 2025-07-16
[5.1 cherry] Prefer integrity_brouting for avoid_unpaved+avoid_footways

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/22687

PR #1250, opened 2025-08-22
Try to add suffix Australia-oceania_2.road.obf.zip

RZR-UA:


RZR-UA:
Will be reverted and replaced with join_road_files=no

PR #1252, opened 2025-08-25
Add support for hsb TTS

Korla-tech:
This PR adds support for Upper Sorbian TTS (ISO code: hsb) to the OsmAnd resources repository.

What I changed:

- Added a new TTS resource set for Upper Sorbian under the same directory layout and naming conventions used by existing TTS resources.
- I modified the Czech TTS file, because the grammar is very simmilar.

If I need to change something or you need any extra info feel free to ask.


Korla-tech:
@xmd5a2 @RZR-UA can you take a look at this?

PR #1373, opened 2026-04-21
Apply dynamic "bus-max" speed for PT routes

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/6827


vshcherb:
3 parameters that we don't need for now

PR #1381, opened 2026-05-12
ge1-unit-test-fix

Gateway07:
remove 'Remseck' token from query because Dorfstraße is became FrequentlyUsed word and 'Remseck' is missed in OBF


vshcherb:
Test for  "6 Dorfstraße Remseck am Neckar"

PR #1397, opened 2026-07-05
Add surface=laterite and stop converting surface=clay to dirt

julcnx:
`surface=laterite` was approved via OSM's proposal process ([Proposal:Surface=laterite](https://wiki.openstreetmap.org/wiki/Proposal:Surface%3Dlaterite), 29-0-0) and is documented at [Tag:surface=laterite](https://wiki.openstreetmap.org/wiki/Tag:surface%3Dlaterite).

**Two separate bugs, one fix:**

1. `obf_creation/rendering_types.xml` had no `<type>` registration for `laterite` at all, so it wasn't being retained as a distinct surface value in the compiled map data.
2. The same file silently rewrites `surface=clay` to `surface=dirt` on import (`entity_convert ... from_value="clay" to_value1="dirt"`), so `clay` never survives to reach a routing profile as `clay` in the first place, any speed factor keyed on `clay` would be dead code.

This PR registers both as retained values, then gives them their own speed factors in the bicycle profile's `multiply_factor` table in `routing/routing.xml`: `clay=0.60`, `laterite=0.55`, placed between the existing `dirt`/`mud` (0.50) and `ground`/`earth` (0.75) entries.

**Reasoning for the values:** `mud` has no dry season, so it keeps the harshest existing factor. `clay` and `laterite` are both seasonal, dry and passable for roughly half the year, so both sit above `mud`. `laterite` sits below `clay` because of an added risk `clay` doesn't have: a near-frictionless "black ice" phase right as it turns wet, distinct from clay's more gradual sticky/plastic transition. After sustained rain and traffic churn, laterite converges to the same sticky, plastic state as wet clay, so this is about capturing that earlier transition window, not describing a permanently worse material. These are a starting position based on that reasoning, not measured data, open to a different split if there's a preferred methodology.

**Photos**, for context:

Dry laterite road with blocky desiccation cracking between wheel tracks

*Dry: desiccation cracks into large, blocky plates, distinct from clay's smaller, curved cracking pattern.* [Source](https://commons.wikimedia.org/wiki/File:Dry_laterite_road_with_blocky_desiccation_cracking_between_wheel_tracks.jpg)

Wet laterite road surface during rainfall, appearing firm

*Wet, phase 1 (near-frictionless): surface still looks firm, but this is the near-frictionless "black ice" phase specific to laterite, distinct from wet clay's visibly sticky, plastic transition.* [Source](https://commons.wikimedia.org/wiki/File:Wet_laterite_road_surface_during_rainfall,_appearing_firm.jpg)

Red clay sticky and plastic when wet

*Wet, phase 2 (after sustained rain and traffic): confirmed laterite despite the filename (mislabeled before `surface=laterite` existed as a distinct tag), showing the later sticky, plastic, heavily rutted state that laterite converges to, the same state wet clay reaches.* [Source](https://commons.wikimedia.org/wiki/File:Red-clay-sticky-plastic-when-wet.jpg)

### DIFF
--- a/obf_creation/rendering_types.xml
+++ b/obf_creation/rendering_types.xml
@@ -2753,7 +2753,8 @@
 		<entity_convert pattern="tag_transform" from_tag="surface" from_value="soil" to_tag1="surface" to_value1="ground"/>
 		<type tag="surface" value="dirt" minzoom="14" additional="true"/>
 		<entity_convert pattern="tag_transform" from_tag="surface" from_value="dirt/sand" to_tag1="surface" to_value1="dirt"/>
-		<entity_convert pattern="tag_transform" from_tag="surface" from_value="clay" to_tag1="surface" to_value1="dirt" poi="false"/>
+		<type tag="surface" value="clay" minzoom="14" additional="true"/>
+		<type tag="surface" value="laterite" minzoom="14" additional="true"/>
 		<type tag="surface" value="mud" minzoom="14" additional="true"/>
 		<type tag="surface" value="ice" minzoom="14" additional="true"/>
 		<type tag="surface" value="salt" minzoom="14" additional="true"/>
--- a/routing/routing.xml
+++ b/routing/routing.xml
@@ -1845,6 +1845,8 @@
 			<select value="0.65" t="surface" v="grass"/>
 			<select value="0.75" t="surface" v="ground"/>
 			<select value="0.5" t="surface" v="mud"/>
+			<select value="0.55" t="surface" v="laterite"/>
+			<select value="0.60" t="surface" v="clay"/>
 			<select value="0.5" t="smoothness" v="very_bad"/>
 			<select value="1.2" t="route_bicycle"/> 
 			<!-- If a part of a bicycle route has a surface tag, it will have


xmd5a2:
Thanks!

aceman444:
Thanks.
But don't forget to check how this turns up in the UI, if the surface name is shown. There are currently 2 strings for clay in Osmand, poi_resource_clay and poi_surface_clay. I guess for roads poi_surface_clay is used, however, as you know 'clay' has different meaning for tennis pitches and roads, thus also different translations. I suspect poi_surface_clay is used for both highways and pitches in OsmAnd.

julcnx:
> There are currently 2 strings for clay in Osmand, poi_resource_clay and poi_surface_clay. [...] I suspect poi_surface_clay is used for both highways and pitches in OsmAnd.

Checked this against `poi/poi_types.xml` and `OsmAnd/res/values/phrases.xml`:

- `laterite` doesn't have a tennis-court or other sport-surface meaning, so there's no equivalent ambiguity there. `poi_surface_laterite` ("Surface: laterite") already exists in `phrases.xml`, it looks like it was auto-added from this PR shortly after it opened, so display should just work once merged, nothing to do on that side.
- You're right about `clay` though: `poi_surface_clay` ("Surface: clay") is a single shared string, driven by one `poi_additional` entry (`surface_clay`, tag=`surface` value=`clay`) applied wherever the generic "surface" attribute group appears, covering both highway/path objects and sport pitches. That said, this entry predates this PR, it's not something this change adds or touches. This PR only edits `rendering_types.xml` (retain the value in the compiled map) and `routing.xml` (speed factor), not `poi_types.xml`, so the translation side is unaffected either way, for better or worse.

So the clay/tennis-pitch ambiguity is a real, pre-existing issue, just independent of this PR. If it's worth fixing on its own, it'd need a second, surface-specific string (e.g. distinguishing sport pitch clay from road/track clay), let me know if you'd like me to look into that separately.

aceman444:
Yes, the string is pre-existing in OsmAnd, what I mean is that now it may be exposed at new places, because you stopped converting clay to dirt (thus string for clay was never used, at least in some cases).
Create a route for pedestrians going over a clay footway, build the route in OsmAnd, then click details, there is a block about 'surface', expand it. You will see list of surface names and metres how long the route follows such surfaces. Does OsmAnd now after your path show clay in that list?
In OsmAnd 5.3.10 (without your change), it actually displays the string rendering_attr_surface_dirt_name. Will there be  rendering_attr_surface_clay_name created after your patch or what does OsmAnd display there?

julcnx:
@aceman444

> In OsmAnd 5.3.10 (without your change), it actually displays the string rendering_attr_surface_dirt_name. Will there be rendering_attr_surface_clay_name created after your patch or what does OsmAnd display there?

Good catch, there was a real gap. The route-details surface breakdown is driven by `routeInfo_surface` in `default.render.xml`, and that got a same-day follow-up commit ([711b568](https://github.com/osmandapp/OsmAnd-resources/commit/711b568)) adding `attrStringValue="surface_clay"` / `"surface_laterite"` cases, so the rendering side is wired up correctly.

But that resolves to `rendering_attr_surface_clay_name` / `rendering_attr_surface_laterite_name` string resources over in the main app repo, and those were never added, unlike every other surface value in that list (`dirt`, `mud`, `ice`, etc. all have one). So until now, a clay or laterite route segment would hit a missing string resource in that list instead of a clean label.

Opened a fix: [osmandapp/OsmAnd#25469](https://github.com/osmandapp/OsmAnd/pull/25469).

aceman444:
Thank you, I can confirm the strings rendering_attr_surface_clay_name and rendering_attr_surface_laterite_name now appear in the list of strings to translate.

PR #1425, opened 2026-08-14
Split Hungary (https://github.com/osmandapp/OsmAnd/issues/25484)

xmd5a2:


aceman444:
Which relations in OSM represent these regions so we can add the translated name:<lang> tags to them?

PR #1428, opened 2026-08-20
fix: string too large to encode using UTF-8

mohe2015:
Happens when building OsmAnd. Fixed by splitting up the path into multiple parts.


dmpr0:
Hi! Thanks for the fix, but in this case, the original SVG should be fixed, because the next time resources are regenerated, the XML will be regenerated too. But this file is no longer used so it should be removed. https://github.com/osmandapp/OsmAnd-resources/commit/b225bb2cf9aa714d0a40ca608e0452e48e9a38db

PR #1436, opened 2026-08-26
bbox31_poi_filtered

Gateway07:
- add bbox31 filtered by 2km


vshcherb:
Not needed as we are not going to use POI bbox for search internal

PR #1448, opened 2026-09-03
Nautical + Marine: label seamarks with seamark:name only (fixes #25570)

### LINKED ISSUE #25570: Nautical rendering style displays unnecessary and wrong strings for buildings
### Description

Nautical rendering style displays unnecessary and wrong strings for buildings.

### Steps to reproduce

Unfortunately, I failed to upload a screenshot for hours.
See e.g. 47° 42' 18.1" North and 11° 44' 25.3" East. It displays "Wasserwacht Bad Wiessee" at house number 15.

### Actual result

The nautical map style renders for a building in blue
1. name
2. operator
3. house number.

### Expected result

It should render
if set seamark:name,
else if set name,
else if set operator:short
else if set operator.

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version: OsmAnd+ 5.3.10
Android version: 14
Device model: Asus Zenfone 9
Crash-Logs: ?
```

vshcherb:
Fixes osmandapp/OsmAnd#25570

## Problem

OsmAnd evaluates the `<text>` rules **once per "name" tag** an object carries
(`TextRenderer.renderText` / the caption loop in `MapPrimitiviser_P.cpp`, with the
`nameTag` input set to that tag). A text rule that does not specify `nameTag`
therefore matches *every* text tag of the object and draws a separate label for
each one.

Almost all of the seamark text rules in `nautical.render.xml` and
`marine.render.xml` were missing `nameTag`. The reported building —
[way/1185270521](https://www.openstreetmap.org/way/1185270521),
`seamark:type=rescue_station` — is stored in the OBF with four name tags:

```
name[<name>]           = Wasserwacht Bad Wiessee
name[seamark:name]     = Wasserwacht Bad Wiessee
name[operator]         = Bayerisches Rotes Kreuz - Ortsgruppe Bad Wiessee
name[addr:housenumber] = 15
```

and all of them were rendered, stacked in blue, exactly as in the report.

Evaluating the text ruleset for `seamark:type=rescue_station` at z18 **before**:

| nameTag | drawn |
|---|---|
| `` (name) | yes |
| `seamark:name` | yes |
| `operator` | yes |
| `addr:housenumber` | yes |
| `ref` | yes |
| `alt_name` | yes |

**after** only `seamark:name` is drawn.

It is not limited to that one object. Sweeping every `tag`/`value` pair of a style
against `{operator, addr:housenumber, alt_name}` and counting the pairs where all
three are drawn (i.e. `nameTag` is genuinely absent, rather than a deliberate
`nameTag="ref"`):

| style | rules without `nameTag` |
|---|---|
| **nautical** | **38** |
| **marine** | **49** |
| osm-carto | 13 |
| topo | 8 |
| default | 6 |
| skimap | 4 |
| desert / UniRS / mapnik / snowmobile | 1–2 |
| LightRS / offroad / Touring-view | 0 |

The nautical family is an order of magnitude worse, and it is where the bug is
actually visible: seamarks routinely sit on buildings and piers that carry
addresses, operators and relation-derived tags. The handful of cases in the other
styles (`natural=land`, `natural=cliff`, `admin_level=2`, …) are latent — those
objects rarely carry such tags. Only the two nautical styles are fixed here.

## Fix

Add the missing `nameTag` to those rules:

* `nameTag="seamark:name"` for `seamark:type=*`, and for the `man_made=*` cases of
  the light block (whose companion `nameTag2` is `seamark:light:character`). This
  is safe and loses nothing: `obf_creation/rendering_types.xml` already copies
  `name` → `seamark:name` for every object with a `seamark:type`
  (`<entity_convert … from_tag="name" if_tag1="seamark:type" if_not_tag2="seamark:name" …>`,
  added in #965). Verified on two generated OBFs (Bad Wiessee, Helgoland): 155
  seamark objects with names, **0** of them have `name` without `seamark:name`.
* `nameTag=""` (the plain/localized name) for the non-seamark rules —
  `leisure=marina`, `waterway=dock`, `waterway=dam`, `waterway=weir`,
  `waterway=waterfall`, `waterway=fairway`, `waterway=flowline`, `route=ferry`,
  `leisure=slipway`, `barrier=city_wall`.
* In `marine`, rocks / wrecks / obstructions pair `nameTag2="depth"` with a
  caption, and an unnamed rock has `depth` as its *only* caption — so each of
  those rules gets a companion case with `nameTag="depth"` (the same two-case
  idiom the file already uses for `restricted_area`). Without it a sounding such
  as "4.7" would disappear; caught by the before/after render below.

The `lightsector=arc|limit|orientation` rules are deliberately left unchanged in
both styles — that data is generated (`notosm`) and there is no sample of it here
to confirm which tag carries its caption.

After the change the sweep reports 3 pairs per style, all of them those
`lightsector` rules.

## Verification

Rendered with the native renderer over two purpose-built OBFs (Overpass extract →
`IndexCreator`), before vs. after at identical lat/lon/zoom, z12–z18. Every
difference found was inspected; all of them are labels that should not have been
there:

* `Adler&Eils` — a ferry `operator` — drawn over the route name (after: the real
  route names `Sylt-Amrum-Helgoland` / `Wilhelmshaven-Hooksiel-Helgoland` become
  legible).
* `238.0-270.0-white-40.0-238.0-red` — `seamark:light:add_sectors` — drawn as its
  own label next to `Binnenhafen, West-Feuer`.
* `Q2` over `Sender Helgoland (F)`, and a bare `F` where the intended `(F)`
  (`seamark:light:character` as `nameTag2`) is already shown.
* `9` on an unnamed `man_made=tower` — the object's `height` tag.
* In Bad Wiessee, `hiking_route_name` ("Europäischer Fernwanderweg E4 …",
  inherited from a hiking relation on the same way) drawn along the ferry line.

Side-by-side images are in the first comment below.

The reported rescue station itself does not change in these images: the legacy
renderer drops the extra labels through text collision at any zoom (they share one
anchor point — checked up to z22), while the OpenGL engine used in the report
stacks them. The rule-evaluation table above is the direct evidence for that
object.

## Not covered here

The issue asks for the fallback chain `seamark:name` → `name` → `operator:short` →
`operator`. The first two are covered (they hold the same value in the OBF). The
`operator` fallback cannot be expressed in a rendering style — a rule cannot test
for the *absence* of another tag — and `operator:short` is not encoded into the
OBF at all. If it is wanted, it belongs in `obf_creation/rendering_types.xml` as
another `entity_convert` filling `seamark:name` from `operator:short`/`operator`
when no name exists, which would need a map rebuild.

Worth considering separately: the engine default ("no `nameTag`" ⇒ "draw every
text tag") is what makes this class of bug possible, and in all 118 rules found
across the styles above it is never what the author intended. Changing the default
to mean `nameTag=""` would fix the remaining latent cases in one place, but that
is an app change and needs its own regression pass.

## AI disclaimer

Produced with Claude Code (Opus 5).

Prompts used (summarised):
1. "Let's test and fix osmandapp/OsmAnd#25570, it looks like a simple bug in the nautical style."
2. "Can't you zoom in? Let's discuss — maybe `operator` should not go into the name section at all, or is it better to fix all the types the way the default style does?"
3. "Let's fix only the two styles then, marine and nautical."

Decided by the agent, not requested explicitly:
- Choosing `nameTag="seamark:name"` (rather than `nameTag=""`) for seamark rules,
  after checking that `rendering_types.xml` guarantees `seamark:name` is present
  whenever `name` is.
- Widening the fix from the reported `seamark:type=rescue_station` rule to all
  sibling text rules in both styles that had the same missing `nameTag`.
- Adding the `nameTag="depth"` companion cases in `marine` to keep soundings on
  unnamed rocks/wrecks/obstructions.
- Leaving the `lightsector=*` rules untouched, as explained above.
- Building throwaway OBFs from Overpass extracts and rendering before/after images
  with the legacy native renderer to verify the change.

### DIFF
--- a/rendering_styles/marine.render.xml
+++ b/rendering_styles/marine.render.xml
@@ -1555,8 +1555,8 @@
 				<case minzoom="17" tag="man_made" value="breakwater" nameTag="" textWrapWidth="20" textOnPath="true" textOrder="255"/>
 				<case minzoom="17" tag="man_made" value="groyne" nameTag="" textWrapWidth="20" textOnPath="true" textOrder="255"/>
 				<switch minzoom="15">
-					<case tag="waterway" value="waterfall" nameTag2="height"/>
-					<case tag="seamark:type" value="waterfall" nameTag2="height"/>
+					<case tag="waterway" value="waterfall" nameTag="" nameTag2="height"/>
+					<case tag="seamark:type" value="waterfall" nameTag="seamark:name" nameTag2="height"/>
 					<apply textWrapWidth="20" textOnPath="true" textOrder="255"/>
 				</switch>
 				<case minzoom="17" tag="seamark:waterway" value="rapids" nameTag="" textSize="11" textOnPath="true"/>
@@ -1596,13 +1596,17 @@
 				An appropriate graphical display of sector lights with coloured circle sections requires further development of code in OSMAND. Feedback welcome: josail @ gmx.de -->
 
 				<switch minzoom="12" additional="depth:exposition=shoaler" textColor="$depthTextColor" textSize="14" textBold="true" textOrder="128" intersectionMargin="10" textDy="-12" nameTag2="depth">
-					<case tag="seamark:type" value="rock"/>
-					<case tag="seamark:type" value="wreck" textDy="-10"/>
+					<case tag="seamark:type" value="rock" nameTag="seamark:name"/>
+					<case tag="seamark:type" value="rock" nameTag="depth"/>
+					<case tag="seamark:type" value="wreck" nameTag="seamark:name" textDy="-10"/>
+					<case tag="seamark:type" value="wreck" nameTag="depth" textDy="-10"/>
 					<apply_if additional="seamark:lnam" textOrder="20"/>
 				</switch>
 				<switch minzoom="13" textColor="$depthTextColor" textSize="14" textItalic="true" textOrder="128" intersectionMargin="10" textDy="-12" nameTag2="depth">
-					<case tag="seamark:type" value="rock"/>
-					<case tag="seamark:type" value="wreck" textDy="-10"/>
+					<case tag="seamark:type" value="rock" nameTag="seamark:name"/>
+					<case tag="seamark:type" value="rock" nameTag="depth"/>
+					<case tag="seamark:type" value="wreck" nameTag="seamark:name" textDy="-10"/>
+					<case tag="seamark:type" value="wreck" nameTag="depth" textDy="-10"/>
 					<apply_if additional="seamark:lnam" textOrder="25"/>
 				</switch>
 				<switch nameTag="seamark:name" nameTag2="seamark:light:character">
@@ -1645,10 +1649,10 @@
 				<switch> <!-- for big icons -->
 					<switch>
 						<!--            <apply textColor="#292929" textWrapWidth="22" textHaloColor="#ddffffff" textHaloRadius="2"/>-->
-						<case minzoom="10" tag="seamark:type" value="light_major" additional="lightsector=source" textOrder="1" textHaloRadius="3">
+						<case minzoom="10" tag="seamark:type" value="light_major" nameTag="seamark:name" additional="lightsector=source" textOrder="1" textHaloRadius="3">
 							<apply_if minzoom="11" nameTag2="seamark:light:character"/>
 						</case>
-						<case minzoom="11" tag="seamark:type" value="light_minor" additional="lightsector=source" textOrder="2" textHaloRadius="3">
+						<case minzoom="11" tag="seamark:type" value="light_minor" nameTag="seamark:name" additional="lightsector=source" textOrder="2" textHaloRadius="3">
 							<apply_if minzoom="12" nameTag2="seamark:light:character"/>
 						</case>
 						<switch minzoom="12" textOnPath="true">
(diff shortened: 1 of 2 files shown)


vshcherb:
Before/after renders (native renderer, identical lat/lon/zoom, purpose-built OBFs from Overpass extracts).

**Helgoland z17 — full tile.** The only difference on the whole tile is that `Adler&Eils` (the ferry `operator`) and a stray bare `F` (`seamark:light:character` drawn as its own label) are gone. Everything else — `Binnenreede`, `Dünenfähre (Schönwetter)`, `Binnenhafen, West-Feuer (F)`, `B1`, `B8`, `Südhafen`, `Die Seenotretter DGzRS` — is unchanged.

Helgoland z17 before/after

Detail of the same area:

Helgoland z17 detail

**Bad Wiessee z19 — the ferry line.** Before, it was also labelled with `hiking_route_name` ("Europäischer Fernwanderweg E4 …", inherited from a hiking relation on the same way), overlapping its own name. After, only the route name is drawn.

Bad Wiessee ferry before/after

The reported rescue station itself cannot show the defect in these images: the legacy renderer drops the extra labels through text collision, while the OpenGL engine used in the report stacks them. The rule-evaluation table in the PR description is the direct evidence for that object.

(gen by AI)

vshcherb:
Marine style, before/after on the same Helgoland OBF (native renderer, identical lat/lon/zoom).

**z18.** `238.0-270.0-white-40.0-238.0-red` is the light's raw `seamark:light:add_sectors` value drawn as its own caption, over `Adler&Eils` (the ferry `operator`). After: the light shows its icon and the intended `(F)`, and the route name `Sylt-Amrum-Helgoland` becomes legible.

marine z18 before/after

**z14.** `Q2…` on top of `Sender Helgoland (F)` — that is the start of the mast's `wikidata` value. The object (`man_made=mast` + `seamark:type=landmark`) carries seven captions in the OBF: `name`, `seamark:name`, `seamark:light:character`, `wikidata`, `height`, `operator` and `note` ("Do not move! Recreate the housenumber node if desired"), and the rule drew all of them.

marine z14 before/after

**z17.** A `9` on an unnamed `man_made=tower` — that is the object's `height` tag (`name[height] = 9`, `name[fixme] = height estimated.`, no name at all).

marine tower height label before/after

Every difference between the two renders at z12–z18 was inspected and is of this kind. The one case where the first attempt *did* remove something wanted was `seamark:type=rock`/`wreck`/`obstruction`: those pair `nameTag2="depth"` with a caption, and an unnamed rock has `depth` as its only caption, so a sounding like `4.7` vanished. Fixed by giving each of those rules a companion case with `nameTag="depth"`, the same two-case idiom the file already uses for `restricted_area`.

(gen by AI)

vshcherb:
Audited on real maps, as asked: `World_seamarks_2.obf` + `Netherlands_noord-holland` + `Netherlands_zuid-holland` — 5.4M named objects. Method: replay the whole `<text>` ruleset (both style versions, zooms 12/15/17/19) over every object and every caption it carries, and diff the resulting label sets. This found a real defect in the first two commits, now fixed in `36ae161`.

### `nameTag="seamark:name"` was too strong an assumption

The `entity_convert` that copies `name` → `seamark:name` is keyed on the **OSM tag** `seamark:type`. But several OBF types of that name are *derived* from something else:

| OBF type | derived from |
|---|---|
| `seamark:type=fairway` | `waterway=fairway` |
| `harbour`, `landmark`, `production_area`, `small_craft_facility`, `notice` | `seamark:<x>:category` / `:function` |
| the light block's `man_made=lighthouse\|tower\|mast\|offshore_platform` | no `seamark:type` at all |

For those the copy never runs, so `seamark:name` does not exist and the caption vanished: **334 of 1464 fairways** (e.g. [way/525918581](https://www.openstreetmap.org/way/525918581) `waterway=fairway` + `name="'t Joppe"`), 46 small craft facilities, 41 harbours, 127 towers, 70 lighthouses, 22 masts. In total **382 objects in Nautical and 229 in Marine lost their name outright**.

Fixed by giving exactly those rules a companion case with the plain `nameTag`, plus a `nameTag="seamark:name"` companion on the three non-seamark rules that can still carry a chart name (`leisure=marina`, `leisure=slipway`, `route=ferry`). Where both captions exist and are equal only one label is drawn — `MapPrimitiviser_P` drops a text symbol whose content duplicates one already added (`hasTwin`), and the legacy renderer collides them — which is why the style already uses this two-case idiom for `harbour`, `landmark`, `dam` and `cable_submarine`. Cost of the companions: a second label on the 159 marinas/slipways whose two names *differ*, versus today where all their captions are drawn.

After the fix: **3 objects** across all three maps still lose a caption that could be read as a name (one of them has an MMSI number, `002734450`, in its `name` tag).

### Why `seamark:name` and not the plain name

Same audit, 213 300 objects carrying a `seamark:type=*` OBF type:

| | count |
|---|---|
| `name` == `seamark:name` | 35 579 |
| both, different values | 6 399 |
| `name` only, no `seamark:name` | **454** |
| `seamark:name` only, no `name` | **124 073** |
| neither (technical captions only) | 46 795 |

So `nameTag=""` alone would drop 124 073 labels; `nameTag="seamark:name"` alone dropped 454, and the companions above bring that to 3.

### What is still dropped on purpose

Objects that end up with no label at all (the count is per style over all three maps, and *every* one of them previously showed a caption that is not a name):

| Nautical | Marine | what used to be drawn |
|---|---|---|
| 1959 | 7909 | `note`, `fixme`, `wikidata`, `ele`, `route_id`, `building:levels`, light-sector strings |
| 1156 | 6354 | `height` on unnamed waterfalls/weirs |
| 1742 | 5374 | `operator` on an unnamed object |
| 797 | 1195 | `ref` / `ferry_route_ref` on an unnamed ferry line |
| 266 | 284 | only a localized `name:xx` and no `name` |

The first two groups are unambiguously junk (a marina captioned with a paragraph of `note`, a dam captioned "Is it really a dam??"). The `operator` group is the fallback the issue asks for and the `ref` group is arguably useful — neither can be expressed in a rendering style, because a rule cannot test for the *absence* of another tag. If they are wanted, that belongs in `obf_creation/rendering_types.xml`, e.g.

```xml
<entity_convert pattern="tag_transform" from_tag="operator:short" if_tag1="seamark:type"
                if_not_tag2="name" if_not_tag3="seamark:name"
                to_tag1="operator:short" to_tag2="seamark:name"/>
<entity_convert pattern="tag_transform" from_tag="operator" if_tag1="seamark:type"
                if_not_tag2="name" if_not_tag3="seamark:name"
                to_tag1="operator" to_tag2="seamark:name"/>
```

which would implement the requested `seamark:name` → `name` → `operator:short` → `operator` chain exactly, and the same file could also close the derived-type hole above at the source (making the style companions redundant for newly built maps, though they are still needed for maps already on people's devices). Both need a map rebuild, so they are not in this PR.

The last group (localized name only, no `name`) behaves the same in `default.render.xml` and every other style, so it is left as is.

(gen by AI)

